PR #1, opened 2018-04-13
Initial structure and settings management

erayd:


src/background.src.js
```diff
@@ -0,0 +1,155 @@
...
+    var settings = getLocalSettings();
+    browser.runtime
+        .sendNativeMessage(appID, {
+            settings: settings,
+            action: "configure"
+        })
+        .then(
+            // native config request is successful
+            function(response) {
+                settings.version = response.version;
```
maximbaz: will this assignment result in saving version as part of the local settings in the localStorage? Do you need the version in the settings? Maybe just pass it as a separate argument to `handleMessage`, which is where the potential logic might be that handles requests differently based on the version?

src/background.src.js
```diff
@@ -0,0 +1,155 @@
...
+                if (settings.stores.length) {
+                    // there are user-configured stores present
+                    for (var key in settings.stores) {
+                        if (
+                            response.response.storeSettings.hasOwnProperty(key)
```
maximbaz: `response.response` catch the eye, shall we maybe rename that field to `data`, so that both host app and extension refer to this as `response.data` ?

src/background.src.js
```diff
@@ -0,0 +1,155 @@
...
+                            }
+                        }
+                    }
+                } else {
+                    // no user-configured stores, so use the default store
+                    settings.defaultStore.name = "default";
+                    settings.defaultStore.path = response.response.defaultPath;
+                    settings.defaultStore.settings =
+                        response.response.defaultSettings;
+                    settings.stores.default = settings.defaultStore;
```
maximbaz: I think the `settings.defaultStore` should be filled regardless of the number of custom stores defined by user, and only this last line (creating `settings.store.default`) makes sense to make when there are no user-configured stores.

src/background.src.js
```diff
@@ -0,0 +1,155 @@
...
+                settings.version = response.version;
+                if (settings.stores.length) {
+                    // there are user-configured stores present
+                    for (var key in settings.stores) {
+                        if (
+                            response.response.storeSettings.hasOwnProperty(key)
+                        ) {
+                            var storeSettings = settings.stores[key].settings;
+                            if (storeSettings) {
+                                settings.stores[key].settings = JSON.parse(
```
maximbaz: If I understand correctly, this overrides local settings made in the extension with the settings received from `.browserpass.json`, should be the other way around

src/manifest.json
```diff
@@ -0,0 +1,33 @@
+{
+  "key": "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAvlVUvevBvdeIFpvK5Xjcbd/cV8AsMNLg0Y7BmUetSTagjts949Tp12mNmWmIEEaE9Zwmfjl1ownWiclGhsoPSf6x7nP/i0j8yROv6TYibXLhZet9y4vnUMgtCIkb3O5RnuOl0Y+V3XUADwxotmgT1laPUThymJoYnWPv+lwDkYiEopX2Aq2amzRj8aMogNBUbAIkCMxfa9WK3Vm0QTAUdV4ii9WqzbgjHVruQpiFVq99W2U9ddsWNZjOG/36sFREuHw+reulQgblp9FZdaN1Q9X5cGcT5bncQIRB6K3wZYa805gFENc93Wslmzu6aUSEKqqPymlI5ikedaPlXPmlqwIDAQAB",
+  "manifest_version": 2,
+  "name": "Browserpass CE",
+  "description": "Chrome extension for zx2c4's pass (password manager) - Community Edition.",
```
maximbaz: There are also [Firefox manifest.json and host.json](https://github.com/dannyvankooten/browserpass/tree/master/firefox), they are slightly different - add them too, so we can see the folder structure. I think instead of `src/manifest.json` you would have to have `chrome` and `firefox` folders, or something like that.

src/package.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "browserpass-extension",
+    "version": "3.0.0",
+    "description": "Chrome extension for zx2c4's pass (password manager) - Community Edition.",
```
maximbaz: Chrome and Firefox 🙂

src/popup.src.js
```diff
@@ -0,0 +1,24 @@
...
+"use strict";
+
+require("chrome-extension-async");
+
+var settings = null;
+if (typeof browser === "undefined") {
+    var browser = chrome;
+}
+
+browser.runtime.sendMessage({ action: "getSettings" }).then(
```
maximbaz: ```response = await browser.runtime.sendMessage(...)``` 😛

.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+indent_brace_style = K&R
```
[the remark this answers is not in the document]
erayd: It does. However, prettier seems to be ignoring it.

Do you know what the name would be for what prettier is doing?

src/Makefile
```diff
@@ -0,0 +1,15 @@
+.PHONY: deps
+
+all: deps prettier background.js popup.js
```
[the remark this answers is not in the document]
erayd: Good point. I think .PHONY is a one-off though, so:
```Makefile
.PHONY: all deps
```

Not 100% certain on that though.

src/background.src.js
```diff
@@ -0,0 +1,155 @@
+//-------------------------------- Background initialisation --------------------------------//
+"use strict";
+
+require("chrome-extension-async");
+
+if (typeof browser === "undefined") {
+    var browser = chrome;
+}
```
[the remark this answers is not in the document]
erayd: It's to allow the global use of `browser` instead of `chrome`.

src/background.src.js
```diff
@@ -0,0 +1,155 @@
...
+        return;
+    }
+
+    var settings = getLocalSettings();
+    browser.runtime
+        .sendNativeMessage(appID, {
+            settings: settings,
+            action: "configure"
+        })
+        .then(
```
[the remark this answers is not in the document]
erayd: Good point. I included it for the basic promise stuff, but forgot that it could do that as well. I'll change it :-).

src/background.src.js
```diff
@@ -0,0 +1,155 @@
...
+"use strict";
+
+require("chrome-extension-async");
+
+if (typeof browser === "undefined") {
+    var browser = chrome;
+}
+
+// native application id
+var appID = "com.github.browserpass";
```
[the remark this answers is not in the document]
erayd: `com.github.browserpass.native` works. The `browserpass-native` option is a no-go sadly; Chrome doesn't allow hyphens.

src/background.src.js
```diff
@@ -0,0 +1,155 @@
...
+
+// default settings
+var defaultSettings = {
+    gpgPath: null,
+    stores: {}
```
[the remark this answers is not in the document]
erayd: If I declare it here, then it will get saved to local storage, which we don't want. That's why it's set elsewhere.

Do you have an alternative suggestion for dealing with that issue?

src/background.src.js
```diff
@@ -0,0 +1,155 @@
...
+// handle incoming messages
+chrome.runtime.onMessage.addListener(receiveMessage);
+
+//----------------------------------- Function definitions ----------------------------------//
+/**
+ * Get loaded settings
+ *
+ * @since 3.0.0
+ *
+ * @return object Loaded settings
```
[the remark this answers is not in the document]
erayd: Can do.

src/background.src.js
```diff
@@ -0,0 +1,155 @@
...
+ * @since 3.0.0
+ *
+ * @param mixed            message      Incoming message
+ * @param MessageSender    sender       Message sender
+ * @param function(mixed)  sendResponse Callback to send response
+ * @return void
+ */
+function receiveMessage(message, sender, sendResponse) {
+    // restrict messages to this extension only
+    if (sender.id !== browser.runtime.id) {
```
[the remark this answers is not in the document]
erayd: If the existing extension doesn't restrict it, then potentially yes. There is nothing stopping other extensions from sending it messages unless you explicitly code the restriction as I've done here.

src/background.src.js
erayd: No. It only saves keys which appear in `defaultSettings`.

Noting it's not saved, is there any reason why `version` should *not* be part of the settings object? It seems a logical place to put it, and simplifies using it elsewhere in the app if we don't need to pass it all over the place via arguments. Tracing callbacks to find forgot-to-pass-something bugs gets annoying pretty fast.

src/background.src.js
erayd: Yep, can do.

src/background.src.js
erayd: OK.

src/background.src.js
erayd: At the moment there are no local settings for stores; I haven't coded that bit yet ;-). But yes agree that, local settings should take priority over `.browserpass.json`.

src/manifest.json
erayd: I haven't got that far yet, but I'm aware they're different. My thought was to have the build process generate `chrome` and `firefox` folders, but have the actual source live in the `src` folder. That way we don't end up with junk in the distributed output (e.g. `*.src.js`, which shouldn't be in the final build).

src/package.json
erayd: Yup. This is the Chrome manifest file though. Should the Chrome description mention Firefox?

src/popup.src.js
erayd: Yup - I'll change it.

.editorconfig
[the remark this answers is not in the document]
maximbaz: I don't, maybe prettier doesn't even support it - maybe let's remove this and just stick with the default prettier style?

src/Makefile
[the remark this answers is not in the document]
maximbaz: Let's have it on every such goal, because it's easier to spot when it's missing 🙂

src/background.src.js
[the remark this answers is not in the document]
maximbaz: I see, I actually realized that settings for the default store are only really useful for "configure" request - after that, you send me the default store together with all of its settings in the regular list of "stores". That's why I created https://github.com/browserpass/browserpass-native/pull/8, look specifically in https://github.com/browserpass/browserpass-native/pull/8/commits/f440e4f5e40bd212c1e0bdd0e71f969ce0a97d22 for the proposal.

src/background.src.js
maximbaz: As long as it's temporarily stored in the settings, not being saved and not being sent to the host app - sure, why not. I simply thought that we will only need version in `handleMessage` and nowhere else, but I can see how this might be wrong.

src/background.src.js
maximbaz: Updated the docs as part of the https://github.com/browserpass/browserpass-native/pull/8 in this commit:
https://github.com/browserpass/browserpass-native/pull/8/commits/d4c82818b8cfeb5421d73afadcead91f6283d823

src/background.src.js
maximbaz: I changed my mind, I don't think we need `settings.defaultStore` at all since you pass the fully described "default" store in the regular list of stores anyway - see my comment https://github.com/browserpass/browserpass-extension/pull/1#discussion_r181476021

src/manifest.json
maximbaz: If we can fully generate `chrome` and `firefox` folders, that would be awesome, but let's add firefox-specific files in this PR so we can agree on where we place them early on in the game.

src/package.json
maximbaz: This is package.json, not part of chrome manifest anymore 🙂 In the manifest let's keep Chrome only.

.editorconfig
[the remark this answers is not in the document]
erayd: Works for me :-).

src/background.src.js
erayd: Agreed.

src/manifest.json
erayd: I reckon generate them. It's easy enough to do, and avoids the need to maintain two copies of what's largely the same thing.

src/package.json
erayd: What about "Browser extension"? That's nice and vendor-neutral ;-).

src/popup.src.js
erayd: This isn't in a function at present, and `await` isn't allowed outside of an `async` scope - I'll do that change in a future PR once this code moves.

erayd:
How's that?

src/Makefile
```diff
@@ -0,0 +1,15 @@
+.PHONY: all
+all: deps prettier background.js popup.js
+
+.PHONY: deps
+deps:
+	yarn install
+
+prettier: *.src.js
+	prettier --write *.src.js
```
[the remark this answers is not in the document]
erayd: Yep, no problem - I'll change this accordingly.

src/background.src.js
```diff
@@ -0,0 +1,152 @@
...
+var appID = "com.github.browserpass.native";
+
+// default settings
+var defaultSettings = {
+    gpgPath: null,
+    stores: {}
+};
+
+// handle incoming messages
+chrome.runtime.onMessage.addListener(receiveMessage);
```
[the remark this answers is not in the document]
erayd: Good catch - thanks :-). Will change.

PR #5, opened 2018-04-19
Port Maxim Baz's form-fill logic

erayd:


src/inject.js
```diff
@@ -1,14 +1,290 @@
...
+                    if (prop.toLowerCase().indexOf(marker) > -1) {
+                        return form;
+                    }
+                }
+            }
+        }
+
+        console.error(
+            "Unable to detect which of the multiple available forms is the login form. Please submit an issue for browserpass on github, and provide the following list in the details: " +
+                JSON.stringify(formProps)
```
maximbaz: With this I wanted to encourage users to submit cases when the auto-fill didn't find a form. Experience shows that people don't look in the console, only one person ever sent this to me 🙂

Let's keep this log entry, but let's keep in mind for future that perhaps in general we might want to show some more visible message to a user, saying something like "we couldn't identify the login form with confidence, consider sending us this link via github issue"

PR #6, opened 2018-04-19
Implement basic auth

erayd:


src/background.js
```diff
@@ -44,6 +44,70 @@ function getLocalSettings() {
...
+/**
+ * Handle modal authentication requests (e.g. HTTP basic)
+ *
+ * @since 3.0.0
+ *
+ * @param object requestDetails Auth request details
+ * @return object Authentication credentials or {}
+ */
+function handleModalAuth(requestDetails) {
+    console.log(this, requestDetails);
```
maximbaz: remove before merge

src/background.js
```diff
@@ -95,6 +159,29 @@ async function handleMessage(settings, message, sendResponse) {
                 if (!url.match(/:\/\//)) {
                     url = "http://" + url;
                 }
+                var modalAuthHandler = handleModalAuth.bind({ url: url, login: message.login });
+                chrome.webRequest.onAuthRequired.addListener(
```
maximbaz: let's get rid of the `browser` alias, I see it is more intuitive to write `chrome` so let's write `chrome` everywhere 🙂

src/background.js
```diff
@@ -95,6 +159,29 @@ async function handleMessage(settings, message, sendResponse) {
...
+                    // ignore non-complete status
+                    if (info.status !== "complete") {
+                        return;
+                    }
+
+                    // remove auth handler once page is loaded
+                    chrome.webRequest.onAuthRequired.removeListener(modalAuthHandler);
+
+                    // remove updated handler
+                    chrome.tabs.onUpdated.removeListener(self);
```
maximbaz: I feel somewhat better about a single listener than dynamically subscribing and unsubscribing to tabs.onUpdated. Especially because it is a global listener on all tabs.

What do you think about the approach in the original browserpass, where we subscribe to `tabs.onUpdated` once in the top of the file, and here we just register the `modalAuthHandler` and save it into a global variable that `onUpdatedListener` will use?

src/popup/popup.js
```diff
@@ -124,6 +125,12 @@ async function withLogin(action) {
                 handleError("Filling login details...", "notice");
                 break;
             case "launch":
+                var havePermission = await chrome.permissions.request({
+                    origins: ["http://*/*", "https://*/*"]
+                });
```
maximbaz: wow, I haven't heard of `optional_permissions` and this approach being a thing!

Can we also request `webRequest` and `webRequestBlocking` here instead of in manifest? So far we only need them for basic auth, which only works if we use the `launch` action.

src/background.js
erayd: I agree. I keep on doing this, despite meaning to use `browser` here... so let's just go with `chrome`.

src/background.js
erayd: I did this because you requested no global state. But if you're happy to have global state (i.e. the list of listeners), then we can use that approach.

src/popup/popup.js
erayd: We can.

src/background.js
maximbaz: I see, I'm primarily against a global state between host app and extension (because the host app can change while browser is still running), but a state during the life-time of the browser extension that doesn't affect anything outside is fine for me 🙂

src/background.js
```diff
@@ -16,8 +12,24 @@ var defaultSettings = {
...
+chrome.tabs.onUpdated.addListener(function(tab, info) {
+    // ignore non-complete status
+    if (info.status !== "complete") {
+        return;
+    }
+
+    // unregister any auth listeners for this tab
+    if (authListeners[tab.id]) {
+        chrome.tabs.onAuthRequired.removeListener(authListeners[tab.id]);
+        authListeners[tab.id] = null;
```
maximbaz: I think this will leave the key in the object, but `delete authListeners[id]` will completely erase it.

src/background.js
```diff
@@ -16,8 +12,24 @@ var defaultSettings = {
     stores: {}
 };
 
+var authListeners = {};
+
+// unregister auth handlers
```
maximbaz: I'd remove the comment on the function level (because we will reuse this function again and forget to change this comment), but keep this comment on the actual code block that is unregistering the auth handler.

Maybe add a more generic comment, such as "handle tab updates", or something like that

src/background.js
```diff
@@ -90,11 +164,21 @@ async function handleMessage(settings, message, sendResponse) {
...
             try {
-                var tab = (await browser.tabs.query({ active: true, currentWindow: true }))[0];
+                var tab = (await chrome.tabs.query({ active: true, currentWindow: true }))[0];
                 var url = message.login.fields.url ? message.login.fields.url : response.login.url;
                 if (!url.match(/:\/\//)) {
                     url = "http://" + url;
                 }
+                if (authListeners[tab.id]) {
+                    chrome.tabs.onUpdated.removeListener(authListeners[tab.id]);
+                    authListeners[tab.id] = null;
```
maximbaz: same idea about using `delete`

erayd:
Rebased on master.

PR #7, opened 2018-04-19
Add button to remove domain filter

erayd:
image

### DIFF
--- a/src/popup/searchinterface.js
+++ b/src/popup/searchinterface.js
@@ -60,10 +60,7 @@ function view(ctl, params) {
                               );
                               target.focus();
                               self.popup.active = false;
-                              self.popup.search(
-                                  target.value.trim(),
-                                  target.value.substr(0, 1) !== " "
-                              );
+                              self.popup.search(target.value);
                           }
                       })
                   ])
@@ -75,7 +72,7 @@ function view(ctl, params) {
                     e.dom.focus();
                 },
                 oninput: function(e) {
-                    self.popup.search(e.target.value.trim(), e.target.value.substr(0, 1) !== " ");
+                    self.popup.search(e.target.value);
                 },
                 onkeydown: function(e) {
                     switch (e.code) {
--- a/src/popup/searchinterface.js
+++ b/src/popup/searchinterface.js
@@ -83,7 +83,7 @@ function view(ctl, params) {
                                     self.popup.search("");
                                 } else if (
                                     e.target.selectionStart == 0 &&
-                                    e.target.selectionStart == e.target.selectionEnd
+                                    e.target.selectionEnd == 0
                                 ) {
                                     self.popup.active = false;
                                     self.popup.search(e.target.value);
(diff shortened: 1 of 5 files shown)


src/popup/searchinterface.js
```diff
@@ -50,7 +50,24 @@ function view(ctl, params) {
...
+            this.popup.active
+                ? m("div.hint.badge", [
+                      this.popup.settings.host,
+                      m("div.remove-hint", {
+                          onclick: function(e) {
+                              var target = document.querySelector(
+                                  ".part.search > input[type=text]"
+                              );
+                              target.focus();
+                              self.popup.active = false;
```
maximbaz: What does `active` mean for the popup? Is it referring exclusively to the fact that we have a domain filter? If yes, can we have a more descriptive name?

src/popup/searchinterface.js
```diff
@@ -50,7 +50,24 @@ function view(ctl, params) {
...
+                              self.popup.search(
+                                  target.value.trim(),
+                                  target.value.substr(0, 1) !== " "
```
maximbaz: Since it's not the first time when you write this logic (trim the first argument, substring the second), can you move it into the `search` function itself? Let it just accept the `target.value` as the only parameter, and do the trimming and comparison inside.

src/popup/searchinterface.js
```diff
@@ -63,9 +80,17 @@ function view(ctl, params) {
...
-                            if (self.popup.active && e.target.value.length == 0) {
-                                self.popup.active = false;
-                                self.popup.search("");
+                            if (self.popup.active) {
+                                if (e.target.value.length == 0) {
+                                    self.popup.active = false;
+                                    self.popup.search("");
+                                } else if (
+                                    e.target.selectionStart == 0 &&
+                                    e.target.selectionStart == e.target.selectionEnd
```
maximbaz: It would be easier to read if `if (e.target.selectionStart == 0 && e.target.selectionEnd == 0)`

src/popup/searchinterface.js
```diff
@@ -63,9 +80,17 @@ function view(ctl, params) {
...
+                                ) {
+                                    self.popup.active = false;
+                                    self.popup.search(e.target.value);
```
maximbaz: I think it should have been here `search(e.target.value, e.target.value.substr(0, 1) !== " ")`, one more supportive case to move this logic inside of `search()` 😛

src/popup/searchinterface.js
erayd: It means "only show logins which are `.active=true`, which means "this login is for the currently active domain". I'll rename it to something more descriptive.

src/popup/searchinterface.js
erayd: Yes.

src/popup/searchinterface.js
erayd: Good point, will change.

src/popup/searchinterface.js
maximbaz: Thanks, `currentDomainOnly` makes more sense to me :)

PR #8, opened 2018-04-21
Add copy actions

erayd:


erayd:
Rebased on master.

src/popup/interface.js
```diff
@@ -88,6 +88,13 @@ function view(ctl, params) {
                                 case "Enter":
                                     result.doAction("fill");
                                     break;
+                                case "KeyC":
+                                    if (e.ctrlKey) {
+                                        result.doAction(
+                                            e.shiftKey ? "copyUsername" : "copyPassword"
```
maximbaz: In the original browserpass extension copy username was `Shift+C`, not `Ctrl+Shift+C`. I don't mind changing if you prefer this one (I actually like `Ctrl+Shift+C` more), just want to make sure it is a conscious decision on your side too.

src/background.js
```diff
@@ -37,6 +37,27 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse) {
...
+function copyToClipboard(text) {
+    document.addEventListener(
+        "copy",
+        function(e) {
+            e.clipboardData.setData("text/plain", text);
+            e.preventDefault();
+        },
+        { once: true }
+    );
+    document.execCommand("copy");
```
maximbaz: This looks too easy to be true 😃 Very interesting approach!

Just keep in mind, Firefox is special, it doesn't allow copying from background page, so we have to also send it to the popup and copy from the popup. I added this to the TODO document in case you want to confirm and deal with it separately.

See https://github.com/browserpass/browserpass/issues/241#issuecomment-376868021

PR #9, opened 2018-04-21
Track recently used logins & display first in popup

erayd:


erayd:
Rebased on master.

erayd:
Rebased again.

src/popup/popup.js
```diff
@@ -80,20 +80,33 @@ async function run(settings) {
...
                     login: response[store][key].replace(/\.gpg$/i, ""),
                     allowFill: true
+                    recent: -1
                 };
                 login.domain = pathToDomain(login.store + "/" + login.login);
                 login.inCurrentDomain =
                     settings.host == login.domain || settings.host.endsWith("." + login.domain);
+                var recent = localStorage.getItem("recent:" + settings.host);
+                if (recent) {
+                    recent = JSON.parse(recent);
```
maximbaz: Would be cool to retrieve and parse the data once, outside of `for (var store in response)` loop, and just use it here. 

Even better would be if the `recent` was a map instead of list, so that we don't have to iterate it every time, but instead use something like `if (recent[storePath] and recent[storePath][login.login]) { ... }`, but I'm not sure how you feel about it

src/popup/popup.js
```diff
@@ -105,6 +118,36 @@ async function run(settings) {
...
+ * Save login to recent list for current domain
+ *
+ * @since 3.0.0
+ *
+ * @param object settings Settings object
+ * @param object login    Login object
+ * @param bool   remove   Remove this item from recent history
+ * @return void
+ */
+function saveRecent(settings, login, remove = false) {
```
maximbaz: `remove` is not implemented, right? Will it be? When would you like to remove it?

src/popup/popup.js
```diff
@@ -80,20 +80,33 @@ async function run(settings) {
...
         var index = 0;
         for (var store in response) {
+            var storePath = settings.stores[store].path;
             for (var key in response[store]) {
                 // set login fields
                 var login = {
                     index: index++,
                     store: store,
                     login: response[store][key].replace(/\.gpg$/i, ""),
                     allowFill: true
```
maximbaz: missing comma breaks `make` ;)

src/popup/popup.js
erayd: Agree retrieving `recent` should be outside the loop. Retrieving it every time makes no sense.

Regarding map vs list - will discuss on wire; there are some considerations re order.

src/popup/interface.js
```diff
@@ -148,15 +162,27 @@ function search(s) {
...
+            if (a.store.when != b.store.when) {
+                return b.store.when - a.store.when;
+            }
+            if (a.recent.count != b.recent.count) {
+                return b.recent.count - a.recent.count;
+            }
+            return b.recent.when - a.recent.when;
+        });
+        candidates = recent.concat(
+            candidates.filter(login => login.inCurrentDomain && !login.recent.count)
```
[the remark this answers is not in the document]
erayd: That was old code too - well spotted!

PR #10, opened 2018-04-21
Style main popup view

maximbaz:
Before:

image

After:

image

-----

Main changes:

- Less colors:
    - Search box is white (not... violet?)
    - Inactive search box and the first entry is light-gray
    - focused / hovered entry is gray
    - visually distinguish focus / hover of entries (fill action) and buttons (copy / navigate actions)
- The first entry is only highlighted if no other entry is focused or hovered.
- Alignments, paddings
- No shadows on buttons, with background on hover they look like buttons anyway (and also shadow conflicted with background color 😉)

### DIFF
--- a/src/popup/popup.css
+++ b/src/popup/popup.css
@@ -24,7 +24,7 @@ body {
     box-sizing: border-box;
     display: flex;
     flex-shrink: 0;
-    min-height: 29px;
+    min-height: 22px;
     width: 100%;
 }
 
@@ -39,11 +39,13 @@ body {
 .part.error {
     color: #f00;
     white-space: normal;
+    padding: 7px;
 }
 
 .part.notice {
     color: #090;
     white-space: normal;
+    padding: 7px;
 }
 
 .part.search {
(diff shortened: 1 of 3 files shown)


src/popup/popup.css
```diff
@@ -24,7 +24,7 @@ body {
     box-sizing: border-box;
     display: flex;
     flex-shrink: 0;
-    min-height: 29px;
+    min-height: 22px;
```
erayd: Where does 22px come from? 29 was 6+6 padding, plus 16px content, plus 1px border.

maximbaz: I didn't calculate it, I noticed that there is too much space under the notice / error text in the popup, and it was caused by this number. It didn't seem to have any effect on the popup in the search state 🙂

erayd: Gotcha. This line could probably go completely - it was initially set to stop flexbox squashing things, before I implemented the scrolling. But we have scrolling now, so...

PR #11, opened 2018-04-21
Fix URL bugs

erayd:
- Where no URL is defined
 - Where no URL is defined in the file, but a valid domain is in the path
 - Where the URL contains a protocol

Also fixes getting the username from the path.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -209,9 +209,7 @@ async function handleMessage(settings, message, sendResponse) {
         case "launch":
             try {
                 var tab = (await chrome.tabs.query({ active: true, currentWindow: true }))[0];
-                var url = message.login.fields.url
-                    ? message.login.fields.url
-                    : message.login.domain;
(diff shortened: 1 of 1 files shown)


src/background.js
```diff
@@ -209,7 +209,12 @@ async function handleMessage(settings, message, sendResponse) {
         case "launch":
             try {
                 var tab = (await chrome.tabs.query({ active: true, currentWindow: true }))[0];
-                var url = message.login.fields.url ? message.login.fields.url : response.login.url;
+                var url = message.login.fields.url
+                    ? message.login.fields.url
+                    : message.login.domain;
```
maximbaz: by the way, there is a shorter way to write this is, if you like it:

````javascript
var url = message.login.fields.url || message.login.domain;
```

erayd: Huh... I always thought that returned a boolean. Awesome! Will change it :-).

PR #13, opened 2018-04-21
Sort full list

erayd:


src/popup/interface.js
```diff
@@ -165,19 +165,19 @@ function search(s) {
...
-            if (a.store.when != b.store.when) {
-                return b.store.when - a.store.when;
-            }
-            if (a.recent.count != b.recent.count) {
-                return b.recent.count - a.recent.count;
-            }
-            return b.recent.when - a.recent.when;
-        });
         candidates = recent.concat(
             candidates.filter(login => login.inCurrentDomain && !login.recent.count)
```
maximbaz: It wasn't instantly obvious to me what's going on here, so let's clarify the behavior with a new variable :)

```javascript
var recent = candidates.filter(login => login.recent.count > 0);
var inCurrentDomain = candidates.filter(login => login.inCurrentDomain && !login.recent.count);
candidates = recent.concat(inCurrentDomain);
```

PR #16, opened 2018-04-22
Container stores

erayd:
This implements everything except tracking recent usage of each 'substore'. If you don't like the idea, we can just throw it away.


erayd:
Approach abandoned.

PR #19, opened 2018-04-23
Support login forms in iframe

maximbaz:
We have to inject the script in all frames, because the login form is not necessarily in the top frame. 

For example, we have a "login with national ID" provider in Denmark, and all banks and government sites show their login form in an iframe.

Funny thing, the request for optional permissions is now completely silent, because as per MDN:

> Of this set, the following permissions are granted silently, without a user prompt: activeTab, cookies, idle, webRequest, webRequestBlocking.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -2,7 +2,7 @@ BROWSERIFY := node_modules/.bin/browserify
 PRETTIER := node_modules/.bin/prettier
 
 CLEAN_FILES := js
-PRETTIER_FILES := $(wildcard *.js popup/*.js *.css popup/*.css)
+PRETTIER_FILES := $(wildcard *.js *.json popup/*.js *.css popup/*.css)
 
 .PHONY: all
 all: deps prettier js/background.dist.js js/popup.dist.js js/inject.dist.js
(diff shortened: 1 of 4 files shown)


maximbaz:
A bit off-topic: I noticed that there is an API to remove the optional permissions once we no longer need them. Today the code is requesting `webRequest` on the first usage of "Open URL", and then the permission stays granted forever. Do we want instead to remove the permission once URL is opened, and request it again next time? Keep in mind, with this PR the request is silent, the popup doesn't appear.

maximbaz:
> iframes may be a different domain than the parent, so we should check for this. We can do that in a separate PR though.

Right now I treat this as a "feature", all Danish banks show an iframe with a government login form, which is hosted on another domain. So I have a symlink "bank.com.gpg -> government-id.com.gpg" and use this credential on bank.com 🙂 

Your comment got me thinking, can we make browserpass loop through available iframes and query host app for all available domains? That way I can avoid having "bank.com" symlink, if browserpass can detect that there is a "government-id.com" iframe and add corresponding credentials to the list.

maximbaz:
> Re the optional permissions - they aren't all silent in chrome, at least at the time I tested it.

That's because at the time of testing you also had "origins", they made browser show the popup. Now we don't see the popup, the permission is granted silently 🙂

> I don't see the need to drop the webRequest permission after acquiring it - does doing so actually gain us anything useful, or mitigate a security concern?

I guess no, ok let's keep things as they are now.

maximbaz:
We are definitely sticking to implementation in #22, thus closing this one.

PR #22, opened 2018-07-24
Fill into iframes on primary frame failure

erayd:
## What
 * If unable to fill password in primary frame, re-attempt on all frames
 * Request all permissions at install time, instead of on-demand
 * Prompt the user on foreign origin fill requests

## Why
 * Because filling all frames should not be the default, as it can lead to credental leakage in cases where there is a malicious frame on the same page as a login form (e.g. via an ad).
 * Prompting for an all-sites permission only during some autofills is confusing to the user, and `activeTab` doesn't allow access to other frames (see [here](https://bugs.chromium.org/p/chromium/issues/detail?id=452944)).
 * Autofilling into e.g. an adserver iframe is undesirable, and should be avoided.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -255,22 +255,19 @@ async function handleMessage(settings, message, sendResponse) {
                 var fillFields = JSON.stringify({
                     login: message.login.fields.login,
                     secret: message.login.fields.secret,
-                    origin: tabOrigin
+                    origin: tabOrigin,
+                    autoSubmit: message.login.autoSubmit ? ["login", "secret"] : []
                 });
                 // fill form via injected script
                 var filledFields = await chrome.tabs.executeScript(tab.id, {
-                    code: `window.browserpass.fillLogin(${fillFields}, ${
-                        message.login.autoSubmit ? "true" : "false"
-                    });`
+                    code: `window.browserpass.fillLogin(${fillFields});`
                 });
                 // try again using all available frames if we couldn't fill a password field
                 if (!filledFields[0].includes("secret")) {
                     filledFields = filledFields.concat(
                         await chrome.tabs.executeScript(tab.id, {
                             allFrames: true,
-                            code: `window.browserpass.fillLogin(${fillFields}, ${
-                                message.login.autoSubmit ? "true" : "false"
-                            });`
+                            code: `window.browserpass.fillLogin(${fillFields});`
                         })
                     );
                 }
--- a/src/background.js
+++ b/src/background.js
@@ -345,10 +345,8 @@ async function handleMessage(settings, message, sendResponse) {
 
                 // dispatch initial fill request
                 var filledFields = await fillFields(targetTab, message.login, ["login", "secret"]);
-                if (!filledFields.length) {
-                    throw new Error("Unable to fill credentials");
-                }
 
+                // no need to check filledFields, because fillFields() already throws an error if empty
                 sendResponse({ status: "ok", filledFields: filledFields });
             } catch (e) {
                 try {
--- a/src/background.js
+++ b/src/background.js
@@ -240,7 +240,11 @@ async function handleMessage(settings, message, sendResponse) {
         case "fill":
             try {
                 var tab = (await chrome.tabs.query({ active: true, currentWindow: true }))[0];
-                await chrome.tabs.executeScript(tab.id, { file: "js/inject.dist.js" });
+                await chrome.tabs.executeScript(tab.id, {
+                    allFrames: true,
+                    file: "js/inject.dist.js"
+                });
+                var tabOrigin = new URL(tab.url).origin;
                 // check login fields
                 if (message.login.fields.login === null) {
                     throw new Error("No login is available");
@@ -250,12 +254,23 @@ async function handleMessage(settings, message, sendResponse) {
                 }
                 var fillFields = JSON.stringify({
                     login: message.login.fields.login,
-                    secret: message.login.fields.secret
+                    secret: message.login.fields.secret,
+                    origin: tabOrigin
                 });
                 // fill form via injected script
-                await chrome.tabs.executeScript(tab.id, {
+                var fillResponse = await chrome.tabs.executeScript(tab.id, {
                     code: `window.browserpass.fillLogin(${fillFields});`
                 });
+                if (!fillResponse[0]) {
+                    // try again using all available frames
+                    fillResponse = await chrome.tabs.executeScript(tab.id, {
+                        allFrames: true,
+                        code: `window.browserpass.fillLogin(${fillFields});`
+                    });
+                }
+                if (!fillResponse[0]) {
+                    throw new Error("No fillable forms available");
+                }
                 sendResponse({ status: "ok" });
             } catch (e) {
                 sendResponse({
--- a/src/background.js
+++ b/src/background.js
@@ -88,6 +88,7 @@ async function fillFields(tab, login, fields) {
     var fillRequest = {
         allowForeign: false,
         autoSubmit: login.autoSubmit ? autoSubmitFields : [],
+        submitTarget: login.submitTarget,
         origin: new URL(tab.url).origin,
         login: login,
         fields: fields
@@ -418,6 +419,7 @@ async function parseFields(settings, login) {
     // parse lines
     login.fields = {
         autoSubmit: ["auto-submit", "autosubmit"],
+        submitTarget: ["submit-target", "submittarget"],
         secret: ["secret", "password", "pass"],
         login: ["login", "username", "user", "email"],
         url: ["url", "uri", "website", "site", "link", "launch"]
@@ -459,7 +461,7 @@ async function parseFields(settings, login) {
     }
 
     // move browserpass settings out of fields
-    for (var field of ["autoSubmit"]) {
+    for (var field of ["autoSubmit", "submitTarget"]) {
         login[field] = login.fields[field];
         delete login.fields[field];
     }
--- a/src/background.js
+++ b/src/background.js
@@ -259,14 +259,18 @@ async function handleMessage(settings, message, sendResponse) {
                 });
                 // fill form via injected script
                 var filledFields = await chrome.tabs.executeScript(tab.id, {
-                    code: `window.browserpass.fillLogin(${fillFields});`
+                    code: `window.browserpass.fillLogin(${fillFields}, ${
+                        message.login.autoSubmit ? "true" : "false"
+                    });`
                 });
                 // try again using all available frames if we couldn't fill a password field
                 if (!filledFields[0].includes("secret")) {
                     filledFields = filledFields.concat(
                         await chrome.tabs.executeScript(tab.id, {
                             allFrames: true,
-                            code: `window.browserpass.fillLogin(${fillFields});`
+                            code: `window.browserpass.fillLogin(${fillFields}, ${
+                                message.login.autoSubmit ? "true" : "false"
+                            });`
                         })
                     );
                 }
@@ -343,6 +347,7 @@ async function parseFields(settings, login) {
 
     // parse lines
     login.fields = {
+        autoSubmit: ["auto-submit", "autosubmit"],
         secret: ["secret", "password", "pass"],
         login: ["login", "username", "user", "email"],
         url: ["url", "uri", "website", "site", "link", "launch"]
@@ -360,7 +365,10 @@ async function parseFields(settings, login) {
 
         // assign to fields
         for (var key in login.fields) {
-            if (Array.isArray(login.fields[key]) && login.fields[key].indexOf(parts[0]) >= 0) {
+            if (
+                Array.isArray(login.fields[key]) &&
+                login.fields[key].indexOf(parts[0].toLowerCase()) >= 0
+            ) {
                 login.fields[key] = parts[1];
                 break;
             }
@@ -378,6 +386,12 @@ async function parseFields(settings, login) {
                 login.fields[key] = null;
             }
         }
+    }
+
+    // move browserpass settings out of fields
+    for (var field of ["autoSubmit"]) {
+        login[field] = login.fields[field];
+        delete login.fields[field];
     }
 }
 
--- a/src/background.js
+++ b/src/background.js
@@ -261,19 +261,23 @@ async function handleMessage(settings, message, sendResponse) {
                 var filledFields = await chrome.tabs.executeScript(tab.id, {
                     code: `window.browserpass.fillLogin(${fillFields});`
                 });
-                if (filledFields[0] === false) {
-                    // try again using all available frames
-                    filledFields = await chrome.tabs.executeScript(tab.id, {
-                        allFrames: true,
-                        code: `window.browserpass.fillLogin(${fillFields});`
-                    });
+                // try again using all available frames if we couldn't fill a password field
+                if (!filledFields[0].includes("secret")) {
+                    filledFields = filledFields.concat(
+                        await chrome.tabs.executeScript(tab.id, {
+                            allFrames: true,
+                            code: `window.browserpass.fillLogin(${fillFields});`
+                        })
+                    );
                 }
-                filledFields = filledFields.reduce(function(fields, addFields) {
-                    if (Array.isArray(addFields)) {
-                        return fields.concat(addFields);
-                    }
-                    return fields;
-                }, []);
+                filledFields = filledFields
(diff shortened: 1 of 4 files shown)


src/background.js
```diff
@@ -60,6 +60,78 @@ function copyToClipboard(text) {
...
+        allFrames: true,
+        file: "js/inject.dist.js"
+    });
+
+    // check that required fields are present
+    for (var field of fields) {
+        if (login.fields[field] === null) {
+            throw new Error(`Required field is missing: ${field}`);
+        }
+    }
```
maximbaz: move this check earlier, no need to inject the script altogether if we won't be using it.

src/background.js
```diff
@@ -60,6 +60,78 @@ function copyToClipboard(text) {
...
+        login: login,
+        fields: fields
+    };
+
+    // fill form via injected script
+    var filledFields = await chrome.tabs.executeScript(tab.id, {
+        code: `window.browserpass.fillLogin(${JSON.stringify(fillRequest)});`
+    });
+
+    // try again using same-origin frames if we couldn't fill a password field
```
maximbaz: I'm not sure we should focus on "password field" specifically, it could be that a form contains only username field and it's okay that only username is filled.

Should we instead have here a check that _something_ was filled?

src/background.js
```diff
@@ -60,6 +60,78 @@ function copyToClipboard(text) {
...
+
+    // simplify the list of filled fields
+    filledFields = filledFields
+        .reduce((fields, addFields) => fields.concat(addFields), [])
+        .reduce(function(fields, field) {
+            if (!fields.includes(field)) {
+                fields.push(field);
+            }
+            return fields;
+        }, []);
```
maximbaz: Don't understand yet what's going on here in this simplification and why we need it, and why we return the `filledFields` altogether - is it part of another feature? If so, let's extract it from this PR as well, let the function return nothing for now, not simplify filledFields, not check the length and not return them.

src/background.js
```diff
@@ -239,24 +311,16 @@ async function handleMessage(settings, message, sendResponse) {
...
-                // check login fields
-                if (message.login.fields.login === null) {
-                    throw new Error("No login is available");
-                }
-                if (message.login.fields.secret === null) {
-                    throw new Error("No password is available");
+                // get tab info
+                var targetTab = (await chrome.tabs.query({ active: true, currentWindow: true }))[0];
+
+                // dispatch initial fill request
```
maximbaz: `dispatch fill request`, without "initial" - there could be multiple anyway inside that function :)

src/background.js
```diff
@@ -335,7 +399,10 @@ async function parseFields(settings, login) {
 
         // assign to fields
         for (var key in login.fields) {
-            if (Array.isArray(login.fields[key]) && login.fields[key].indexOf(parts[0]) >= 0) {
+            if (
+                Array.isArray(login.fields[key]) &&
+                login.fields[key].indexOf(parts[0].toLowerCase()) >= 0
```
maximbaz: should `login.fields[key]` be also converted to lower case before searching?

src/inject.js
```diff
@@ -73,24 +73,56 @@
...
+                }
+                var message =
+                    "You have requested to fill sensitive credentials into an embedded document from a " +
+                    "different origin than the main document in this tab. Do you wish to proceed?\n\n" +
+                    `Tab origin: ${request.origin}\n` +
+                    `Embedded origin: ${window.location.origin}`;
+                if (confirm(message)) {
+                    update(PASSWORD_FIELDS, request.login.fields.secret, loginForm);
+                    filledFields.push("secret");
+                }
```
maximbaz: We have to remember the answer, or I will be asked this question all the time

src/inject.js
```diff
@@ -73,24 +73,56 @@
...
+
+        // get the login form
         var loginForm = form();
 
-        update(USERNAME_FIELDS, login.login, loginForm);
-        update(PASSWORD_FIELDS, login.secret, loginForm);
+        // fill login field
+        if (
+            request.fields.includes("login") &&
+            update(USERNAME_FIELDS, request.login.fields.login, loginForm)
```
maximbaz: login is just as sensitive as password, especially if it is an email address 🙂  We should do the same as what you did for password below.

src/inject.js
```diff
@@ -73,24 +73,56 @@
...
             password_inputs[1].select();
         } else {
             window.requestAnimationFrame(function() {
-                // Try to submit the form, or focus on the submit button (based on user settings)
+                // try to locate the submit button
                 var submit = find(SUBMIT_FIELDS, loginForm);
+
+                // Try to submit the form, or focus on the submit button (based on user settings)
                 if (submit) {
                     if (autoSubmit) {
```
maximbaz: `autoSubmit` is now always false, do you expect it to be part of `request`? In this case we can adapt the usage of this variable, i.e. `if (request.autoSubmit)` everywhere

src/popup/popup.js
```diff
@@ -208,11 +201,17 @@ async function withLogin(action) {
...
+                        `Filled fields: ${response.filledFields.sort().join(", ")}`,
+                        "notice"
+                    );
+                // fall through to update recent
                 case "copyPassword":
                 case "copyUsername":
                     saveRecent(this.settings, this.login);
             }
-            window.close();
+            setInterval(() => window.close(), action == "fill" ? 1000 : 0);
```
maximbaz: Why?

src/background.js
erayd: The problem with this is that it ends up with things like search fields etc. being filled with the username, and the iframes being ignored completely.

If there's a way we can address this concern, then I agree that 'anything filled' is a better condition.

src/background.js
erayd: It's used for two things:

1. Reporting to the user which fields were filled (it appears briefly in the popup before the popup closes, and is in this PR)
2. Figuring out which fields still need to be filled, when doing multi-page autosubmit (different PR).

src/background.js
erayd: I'm trying not to change too many context lines - it's relevant when the multi-submit stuff gets applied on top of this. Can we leave this comment as-is?

src/inject.js
erayd: Will add this. Also need a mechanism to clear it.

src/background.js
erayd: Nope, because the entire lookup table is already lowercase.

src/inject.js
erayd: It is initially set from `request.autoSubmit`, but is altered within the function when the autosubmit code is there. It does need to be a separate variable IMO; changing argument objects is a bit of an anti-pattern which annoys me and leads to bugs in the calling scope because the callee changed something.

src/popup/popup.js
erayd: Because the filled fields are displayed to the user briefly before the popup is closed.

src/inject.js
erayd: OK.

src/background.js
erayd: Done.

src/background.js
maximbaz: Could you explain the simplification then, what happens there and why we need it? It's a bit difficult to get the context 🙂  Or move it out as well, now that we agreed not to show the filled fields popup.

src/inject.js
maximbaz: We can leave out the clearing to a different PR, we want to clear out the recent logins for example too

src/background.js
erayd: Each frame returns its own result, so I need to merge all those. Then, after merging, there may be duplicates - so they need removing too. The end result is that `[["login"]["login", "secret", "otp"]]` would become `["login", "secret", "otp"]`.

src/popup/popup.js
```diff
@@ -208,6 +201,12 @@ async function withLogin(action) {
         } else {
             switch (action) {
                 case "fill":
+                    // advise the user which fields were filled & close the popup
+                    handleError(
+                        `Filled fields: ${response.filledFields.sort().join(", ")}`,
+                        "notice"
+                    );
+                // fall through to update recent
```
maximbaz: as discussed, remove this

src/inject.js
maximbaz: do you want to do something in this PR or leave it for future?

src/popup/popup.js
erayd: Per discussion on Wire, have reverted this.

src/inject.js
erayd: Leave it I think. The auto-submit PR already deals with it.

src/popup/popup.js
erayd: Done.

src/background.js
maximbaz: agreed on Wire to try this together with the idea to better select form

src/background.js
erayd: Relating to this - I've been trying a few different options that allow the user to select a form prior to autofilling, and have the selected form be used as the fill target (and then remembered for next time). This is a useful improvement in security, and reduces the risk of filling into unwanted forms, or forms in other frames - but it does add complexity to the behavior, and makes things a bit trickier to reason about.

Do you have an opinion on this either way? I'm undecided - the behavior is useful, and solves a real problem, but I'm not sure whether the added complexity is worth it.

What if the fill was triggered by a visual cue inside the form field (e.g. a Lastpass-like field menu)?

maximbaz: It is a cool idea, but let's leave it out for now and come back to it later. Currently I'm also undecided, back when I was using Lastpass I hated their visual cue as it was quite often interfering with something 🙂

erayd: OK.

src/background.js
```diff
@@ -60,6 +60,107 @@ function copyToClipboard(text) {
...
+    // inject script
+    await chrome.tabs.executeScript(tab.id, {
+        allFrames: true,
+        file: "js/inject.dist.js"
+    });
+
+    // build fill request
+    var fillRequest = {
+        allowForeign: false,
+        allowNoSecret: false,
```
maximbaz: these 2 fields will be added in `dispatchFill` anyway, no need to set them to `false` here

src/background.js
```diff
@@ -60,6 +60,107 @@ function copyToClipboard(text) {
...
+        filledFields = filledFields.concat(await dispatchFill(tab, fillRequest, true));
+    }
+
+    // try again using all available frames if we couldn't fill a password field
+    if (!filledFields.includes("secret")) {
+        filledFields = filledFields.concat(await dispatchFill(tab, fillRequest, true, true));
+    }
+
+    // try again using same-origin frames, and don't require a password field
+    if (!filledFields.length) {
```
maximbaz: shouldn't be `if (!filledFields.includes("secret")) {` here? On the previous step we still tried to fill password

src/background.js
```diff
@@ -239,29 +340,28 @@ async function handleMessage(settings, message, sendResponse) {
...
-                }
-                if (message.login.fields.secret === null) {
-                    throw new Error("No password is available");
+                // get tab info
+                var targetTab = (await chrome.tabs.query({ active: true, currentWindow: true }))[0];
+
+                // dispatch initial fill request
+                var filledFields = await fillFields(targetTab, message.login, ["login", "secret"]);
+                if (!filledFields.length) {
+                    throw new Error("Unable to fill credentials");
```
maximbaz: I think this check is not needed, `fillFields` already throws an Error if it couldn't fill anything.

src/inject.js
```diff
@@ -73,24 +73,58 @@
...
+        if (
+            request.fields.includes("login") &&
+            update(USERNAME_FIELDS, request.login.fields.login, loginForm)
+        ) {
+            filledFields.push("login");
+        }
 
+        // fill secret field
+        if (find(PASSWORD_FIELDS, loginForm)) {
+            update(PASSWORD_FIELDS, request.login.fields.secret, loginForm);
```
maximbaz: probably no need to call `find`, keep the same approach as with login?

```
if (
    request.fields.includes("secret") &&
    update(PASSWORD_FIELDS, request.login.fields.secret, loginForm)
)
```

src/inject.js
```diff
@@ -73,24 +73,58 @@
...
+        }
+
+        // ensure the origin is the same, or ask the user for permissions to continue
+        if (window.location.origin !== request.origin) {
+            var message =
+                "You have requested to fill login credentials into an embedded document from a " +
+                "different origin than the main document in this tab. Do you wish to proceed?\n\n" +
+                `Tab origin: ${request.origin}\n` +
+                `Embedded origin: ${window.location.origin}`;
+            if (!request.allowForeign || !confirm(message)) {
```
maximbaz: The successful confirmation needs to be remembered (in localStorage of background script, or in localstorage of this domain?). The latter is probably fine, and very easy because it doesn't require communication back to background script from here.

maximbaz: It is possible that this dialog will appear twice, easy to repro if you click "No" (it is asked once when we are looking only for passwords, and second time when we are looking for any field). I'm not sure storing negative answer in localStorage is a good idea, what if user clicked "No" by accident... Unless we provide a simple way to reset the answer stored in localStorage.

What do you think?

src/background.js
```diff
@@ -239,29 +340,28 @@ async function handleMessage(settings, message, sendResponse) {
...
+                try {
+                    sendResponse({
+                        status: "error",
+                        message: e.toString()
+                    });
+                } catch (e) {
+                    // TODO An error here is typically a closed message port, due to a popup taking focus
+                    // away from the extension menu and the menu closing as a result. Need to investigate
+                    // whether triggering the extension menu from the background script is possible.
+                    console.log(e);
```
maximbaz: I don't think it is possible to programmatically open browserpass popup. Also, I think the same happens with sending "ok" string, if a confirmation box appeared (which closed popup) and I clicked "yes", credentials were filled, but the credentials were not saved as the most recent ones.

I propose we don't handle this error at all, don't wrap in try-catch, but instead just move all important functionality into the background script. Right now I think the only important functionality that is breaking because of closed message port is updating "recent". Can we update it here, in background script?

This way, in popup we will only have code for showing messages for the user. If popup got closed, well, we won't be able to show the message, but that should be it, the rest of the actions should be correctly performed.

src/background.js
erayd: We don't want to keep them for clarity?

src/background.js
erayd: No, because this step is checking to see whether we failed to fill *anything*, rather than just failing to fill a password.

src/background.js
maximbaz: I'd remove because even if you set them to `true` here, these values aren't respected and get overridden in `dispatchFill` anyway.

src/inject.js
erayd: I think local storage of the background script, because that allows us to clear it from the extension. If we use the local storage of the origin, then we can't get at it unless the user is actually visiting it and our script is injected.

Adding the ability to reset this is important, and IMO belongs in the per-entry menu. Additionally, your argument equally applies to accidentally clicking <kbd>yes</kbd>. Either we should store the state, or we shouldn't, but I don't think we should be privileging one over the other. Thoughts?

maximbaz: Agree on both points, if you want we can implement this in a new PR, this one is too large already anyway

src/background.js
erayd: Yes, we can move this to recent. I'd still like to keep a console message here though, because if we later acquire the ability to do this in a later browser version, I'd like to handle it properly, and this keeps it in mind.

maximbaz: Fair 👍

src/background.js
erayd: Done.

maximbaz:
@erayd hey, ping me on wire 😉

src/inject.js
maximbaz: #29

src/background.js
maximbaz: #30

src/inject.js
maximbaz: don't remember what this was about, let's assume it was handled :)

src/background.js
maximbaz: don't remember what kind of idea for "better select form" we had, but over the past year of using browserpass from this branch I haven't had a single issue with this algorithm.

maximbaz:
Will be addressing the remaining comments in separate PRs, merging this now!

PR #23, opened 2018-07-24
Improve keyboard navigation

erayd:
* Move event handler into separate function;
 * Add left / right arrow navigation;
 * Handle <kbd>TAB</kbd> manually in order to prevent <kbd>SHIFT+TAB</kbd> from wrapping.

### DIFF
--- a/src/popup/interface.js
+++ b/src/popup/interface.js
@@ -67,8 +67,8 @@ function view(ctl, params) {
                         key: result.index,
                         tabindex: 0,
                         onclick: function(e) {
-                            var action = null;
-                            if ((action = e.target.getAttribute("action"))) {
+                            var action = e.target.getAttribute("action");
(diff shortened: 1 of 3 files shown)


src/popup/interface.js
```diff
@@ -67,42 +67,14 @@ function view(ctl, params) {
...
-                                        result.doAction(
-                                            e.shiftKey ? "copyUsername" : "copyPassword"
-                                        );
-                                    }
-                                    break;
-                                case "KeyG":
-                                    result.doAction("launch");
-                                    break;
+                            var action = null;
+                            if ((action = e.target.getAttribute("action"))) {
```
maximbaz: since you have the declaration line anyway, can we change this to:

```
var action = e.target.getAttribute("action");
if (action) {
```

PR #41, opened 2019-03-16
Add Firefox make goal

maximbaz:
@erayd wasn't able to cleanly keep a single manifest for both browsers, if you ever feel like giving it another try, please do 😉

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-/chrome
+/chromium
+/firefox
 /src/node_modules
 /src/css
 /src/js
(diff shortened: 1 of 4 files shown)


src/manifest-chromium.json
```diff
@@ -4,7 +4,7 @@
     "name": "Browserpass",
     "description": "Browser extension for zx2c4's pass (password manager)",
     "version": "3.0.0",
-    "author": ["Maxim Baz <browserpass@maximbaz.com>", "Steve Gilberd <steve@erayd.net>"],
+    "author": "Maxim Baz <browserpass@maximbaz.com>, Steve Gilberd <steve@erayd.net>",
```
maximbaz: Apparently according to Firefox and MDN this can only be a string, changed for Chromium as well

src/manifest-chromium.json
```diff
@@ -23,6 +23,7 @@
         "open_in_tab": false
     },
     "permissions": [
+        "clipboardWrite",
```
maximbaz: this is required on Firefox and optional in Chromium, but added in both places for consistency

.gitignore
```diff
@@ -1,4 +1,5 @@
-/chrome
+/chromium
```
maximbaz: note: renamed to chromium since we support not only chrome, but chromium and its derivatives too 😛

PR #58, opened 2019-04-01
Better handle case when there are multiple forms on the page

maximbaz:
- added `form.action` to the list of attributes, in which a "proper" login form might contain FORM_MARKERS (e.g. a string "login")
- when multiple forms contain FORM_MARKERS, try to further narrow down the list by only selecting the forms that have a password field

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -268,43 +268,47 @@
      * @return The login form
      */
     function form() {
-        var elems = queryAllVisible(document, INPUT_FIELDS, undefined);
-        var forms = [];
-        for (var i = 0; i < elems.length; i++) {
-            var form = elems[i].form;
+        const elems = queryAllVisible(document, INPUT_FIELDS, undefined);
+        const forms = [];
+        for (let elem of elems) {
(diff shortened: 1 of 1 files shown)


src/inject.js
```diff
@@ -268,43 +268,47 @@
...
                 }
             }
         }
 
-        console.error(
-            "Unable to detect which of the multiple available forms is the login form. Please submit an issue for browserpass on github, and provide the following list in the details: " +
-                JSON.stringify(formProps)
-        );
-        return forms[0];
+        // Try to filter only forms that have a password field
```
erayd: How does this handle cases when the request doesn't include `secret`? If we don't want a password form to begin with...

maximbaz: It's just an intuition to find the best form, if we don't want to fill a password, it is likely because we couldn't find a password field in the first place, in which case `formsWithPassword` will be empty and we will fallback to `markedForms` or `forms`.

erayd: Makes sense :+1:

PR #65, opened 2019-04-02
Create a separate icon for popup with exaggerated features

maximbaz:
Looks much better than status quo, even though might not be perfect just yet 🙂 

Both large and small icons are changed.

icon


maximbaz:
@Tobaloidee seeing what a nice logo you've proposed to py3status, I was wondering if I may ask for your opinion on the logo for this project? What you see in the picture above is a new icon that we as a community have come up with for this browser extension. The project is a password manager, tightly coupled with [pass](https://www.passwordstore.org/), therefore the idea for an icon was to have a letter `P` (for "pass") descending into a key - see [this thread](https://github.com/browserpass/browserpass/issues/282) for how it was proposed and implemented.

None of us are really designers, what you see above is the best result we've managed to achieve so far. I do like the idea (although if you have some proposals, feel free to share), our main challenge was to make the icon "recognizable" when it's small and stands next to the address bar in the browser, it would be a pity of majority of people only saw "P" and didn't see any resemblance to a key. It currently consists of a single blue color, because that's the only way I could make the icon look good in browsers with both light and dark theme.

If you have any proposals for how to make it even more "contrast", if you know how to improve the shape of the "P key", or if you have any other feedback, I would really appreciate 😉 

No worries if you don't have the time, I thought it was worth a shot asking you anyway 😉

Tobaloidee:
My pleasure to do logos @maximbaz..earliest tomorrow I will begin designing for this project..👍😃😊

maximbaz:
Thanks! When you get to it, feel free to create a new issue, or maybe just post in https://github.com/browserpass/browserpass/issues/282 so that the rest of the folks will receive notifications 👍

PR #70, opened 2019-04-04
Support OpenID

maximbaz:
OpenID is part of login scenario, therefore we need to support it.

I'm not decided on one particular detail: if a password entry contains "openid" field, and a website has two forms, one asking for username and password, and another one asking for OpenID url, should we fill both forms or only the latter one?

Currently I implemented the former, Browserpass fills both forms (and focuses login button in the OpenID form), because if the OpenID field cannot be detected, we _anyway_ proceed with trying to find and fill a regular form with username and password... but I'm not 100% sure this is the right approach.

@mcnesium, @erayd, thoughts on this?

The example page with two forms: https://www.openstreetmap.org/login (click on OpenID button)

Fixes https://github.com/browserpass/browserpass/issues/208

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -295,7 +295,7 @@ async function fillFields(settings, login, fields) {
     let allowForeign = false;
     let allowNoSecret = !fields.includes("secret");
     let filledFields = [];
-    let importantFieldToFill = fields.includes("secret") ? "secret" : "openid";
+    let importantFieldToFill = fields.includes("openid") ? "openid" : "secret";
 
     // fill form via injected script
     filledFields = filledFields.concat(
--- a/src/background.js
+++ b/src/background.js
@@ -278,13 +278,6 @@ async function dispatchFocusOrSubmit(settings, request, allFrames, allowForeign)
  * @return array List of filled fields
  */
 async function fillFields(settings, login, fields) {
-    // check that required fields are present
-    for (var field of fields) {
-        if (login.fields[field] === null) {
-            throw new Error(`Required field is missing: ${field}`);
-        }
-    }
-
     // inject script
     await chrome.tabs.executeScript(settings.tab.id, {
         allFrames: true,
--- a/src/background.js
+++ b/src/background.js
@@ -298,49 +298,55 @@ async function fillFields(settings, login, fields) {
         fields: fields
     };
 
-    // fill form via injected script
     let allFrames = false;
     let allowForeign = false;
-    let allowNoSecret = false;
-    let filledFields = await dispatchFill(
-        settings,
-        fillRequest,
-        allFrames,
-        allowForeign,
-        allowNoSecret
+    let allowNoSecret = !fields.includes("secret");
+    let filledFields = [];
+    let importantFieldToFill = fields.includes("secret") ? "secret" : "openid";
+
+    // fill form via injected script
+    filledFields = filledFields.concat(
+        await dispatchFill(settings, fillRequest, allFrames, allowForeign, allowNoSecret)
     );
 
-    // try again using same-origin frames if we couldn't fill a password field
-    if (!filledFields.includes("secret")) {
+    // try again using same-origin frames if we couldn't fill an "important" field
+    if (!filledFields.includes(importantFieldToFill)) {
         allFrames = true;
         filledFields = filledFields.concat(
             await dispatchFill(settings, fillRequest, allFrames, allowForeign, allowNoSecret)
         );
     }
 
-    // try again using all available frames if we couldn't fill a password field
-    if (!filledFields.includes("secret") && settings.foreignFills[settings.host] !== false) {
+    // try again using all available frames if we couldn't fill an "important" field
+    if (
+        !filledFields.includes(importantFieldToFill) &&
+        settings.foreignFills[settings.host] !== false
+    ) {
         allowForeign = true;
         filledFields = filledFields.concat(
             await dispatchFill(settings, fillRequest, allFrames, allowForeign, allowNoSecret)
         );
     }
 
-    // try again using same-origin frames, and don't require a password field
-    if (!filledFields.length) {
-        allowForeign = false;
+    // try again, but don't require a password field (if it was required until now)
+    if (!allowNoSecret) {
         allowNoSecret = true;
-        filledFields = filledFields.concat(
-            await dispatchFill(settings, fillRequest, allFrames, allowForeign, allowNoSecret)
-        );
-    }
 
-    // try again using all available frames, and don't require a password field
-    if (!filledFields.length && settings.foreignFills[settings.host] !== false) {
-        allowForeign = true;
-        filledFields = filledFields.concat(
-            await dispatchFill(settings, fillRequest, allFrames, allowForeign, allowNoSecret)
-        );
+        // try again using same-origin frames
+        if (!filledFields.length) {
+            allowForeign = false;
+            filledFields = filledFields.concat(
+                await dispatchFill(settings, fillRequest, allFrames, allowForeign, allowNoSecret)
+            );
+        }
+
+        // try again using all available frames
+        if (!filledFields.length && settings.foreignFills[settings.host] !== false) {
+            allowForeign = true;
+            filledFields = filledFields.concat(
+                await dispatchFill(settings, fillRequest, allFrames, allowForeign, allowNoSecret)
+            );
+        }
     }
 
     if (!filledFields.length) {
@@ -350,7 +356,8 @@ async function fillFields(settings, login, fields) {
     // build focus or submit request
     let focusOrSubmitRequest = {
         origin: new URL(settings.tab.url).origin,
-        autoSubmit: getSetting("autoSubmit", login, settings)
+        autoSubmit: getSetting("autoSubmit", login, settings),
+        filledFields: filledFields
     };
 
     // try to focus or submit form with the settings that were used to fill it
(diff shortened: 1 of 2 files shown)


erayd:
@maximbaz I think we should be filling one or the other, not both. If someone is using OpenID, they are often doing so because they don't trust (or don't want to need to trust) the target website with authentication credentials. If we are filling user / password into that site regardless, we are leaking the user's credentials where they may not want them leaked, even though they have requested a login on the target page.

Also, we should not fill the OpenID field if we are already entering a user / pass. Some sites assume that if the OpenID field has something in it, then they should use OpenID (and not the user / pass), even if other credentials have been entered.

I personally feel that the most appropriate approach is to *never* fill user / pass if an `openid:` field is present in the password entry. If we can find an OpenID field and fill that in, perfect. If we can't, then we should fail safe and not fill anything.

src/inject.js
```diff
@@ -84,7 +86,14 @@
         };
 
         // get the login form
-        var loginForm = form();
+        let loginForm = undefined;
+        if (request.fields.length === 1 && request.fields[0] === "openid") {
```
erayd: Why are we checking the number of array elements here, rather than simply checking whether the request contains the "openid" field?

src/background.js
```diff
@@ -294,20 +290,57 @@ async function fillFields(settings, login, fields) {
...
+                        allFrames,
+                        allowForeign,
+                        allowNoSecret
+                    )
+                );
+            }
+        }
+    }
+
+    // reset request back to filling all fields
```
erayd: Filling all fields when `openid:` is present is a security concern (see previous comment).

src/background.js
```diff
@@ -278,12 +278,8 @@ async function dispatchFocusOrSubmit(settings, request, allFrames, allowForeign)
...
  */
 async function fillFields(settings, login, fields) {
-    // check that required fields are present
-    for (var field of fields) {
-        if (login.fields[field] === null) {
-            throw new Error(`Required field is missing: ${field}`);
-        }
-    }
+    // remove fields that are not present in the login
+    fields = fields.filter(field => field in login.fields);
```
erayd: The null check was explicit in order to detect the difference between a configured (but empty) field, and a nonexistent field. Is there a reason you've changed it to a loosely-typed test?

src/background.js
```diff
@@ -278,12 +278,8 @@ async function dispatchFocusOrSubmit(settings, request, allFrames, allowForeign)
  * @return array List of filled fields
  */
 async function fillFields(settings, login, fields) {
-    // check that required fields are present
-    for (var field of fields) {
-        if (login.fields[field] === null) {
-            throw new Error(`Required field is missing: ${field}`);
```
erayd: Why has the exception been removed?

src/background.js
```diff
@@ -676,7 +710,11 @@ async function handleMessage(settings, message, sendResponse) {
         case "fill":
             try {
                 // dispatch initial fill request
-                var filledFields = await fillFields(settings, message.login, ["login", "secret"]);
+                var filledFields = await fillFields(settings, message.login, [
+                    "login",
```
erayd: IMO these should not occur in the same request. It should be either `["login", "secret"]` or `["openid"]`.

maximbaz:
Interesting idea about filling only OpenID if it is present in a password entry, it seems to make sense on all example websites that I collected for testing.

Here's a particularly good example, [this website](https://ticket.coreboot.org/login) has both password and OpenID _in the same form_, but then if you only fill OpenID (try value `https://id.mcnesium.com`), and submit the form, you are redirected elsewhere to submit ID and password.

@mcnesium would like your opinion as well, in particular because in [an example you posted earlier](https://github.com/browserpass/browserpass/issues/208#issuecomment-364163820) your password entry contains `password`, `username` AND `openid`: was it just an example, and in real life you only have `openid` field password entries for those websites where you want to login with OpenID? Do you agree that if an `openid` is present in the password file, Browserpass should never even attempt to fill `username` and `password` fields, or can you think of a counter-example?

mcnesium:
Thanks a lot for bringing this forward, folks! :+1: 

I totally agree with @erayd in terms of only filling openid, if present in both the login form and the password file.

For now, I never fill OpenID fields using browserpass. If OpenID is supported, I focus the form field and hope my browser has cached an earlier entry, otherwise I just type it.

You are right, [my earlier post](https://github.com/browserpass/browserpass/issues/208#issuecomment-364163820) has just been an example, regarding the structure of the password file. Haven't had the chance to analyse the browserpass code, but I assume my suggestion requires too complicated logic.

Some services still require users to register an account and eventually provide a password, even when using OpenID (e.g. [Gitea](https://try.gitea.io)). So even if not using password login, I would still like to store the password. So the first line of the file must still be reserved for it.

In compliance with [the zx2c4 suggestion](https://www.passwordstore.org/#organization), a dedicated `openid: …` field in the file, that if present is being preferred over user/pass, would be cool for me.

maximbaz:
> Some services still require users to register an account and eventually provide a password, even when using OpenID (e.g. Gitea). So even if not using password login, I would still like to store the password. So the first line of the file must still be reserved for it.

But as per @erayd's suggestion, regardless if a _form_ contains OpenID field, because an `openid:` field is present in your `gitea.io.gpg` file, Browserpass will _never_ attempt to auto-fill password for you, is that correct behavior from your point of view?

mcnesium:
> regardless if a form contains OpenID field, because an openid: field is present in your gitea.io.gpg file, Browserpass will never attempt to auto-fill password for you, is that correct behavior from your point of view?

I guess so, yes. It seems to make sense that if I want to use OpenID to login on a site, I add the `openid:` field to the particular file. If I don't want to use OpenID, then I should not add such a field.

maximbaz:
> Some services still require users to register an account and eventually provide a password

Just to double check that I understand everything correctly (because I'm not using OpenID), so such a password you will store in this password entry, but you will never use it to login, because you will always be loggin in with OpenID, correct?

mcnesium:
That's right.

maximbaz:
Thanks! I'll ping you for testing once I adapt the implementation 😉

src/inject.js
maximbaz: Irrelevant anymore, it was relevant if we were filling both secrets and openid at the same time

src/background.js
maximbaz: Irrelevant anymore, it was relevant if we were filling both secrets and openid at the same time, where some of the fields might have been optional.

src/background.js
maximbaz: Could you please clarify? Based on how our password entry parsing looks now, I don't think the value can ever be `null`. Also, given that this function is only called with either `["login", "secret"]` (which are guaranteed to always be present) or `["openid"]` (which is only the case when "openid" was parsed, so again it's guaranteed to always be present), I don't really understand what this entire loop is trying to catch...

Also see the new line 685, where I added another loosely-typed test, is that an issue?

```javascript
let fields = message.login.fields.openid ? ["openid"] : ["login", "secret"];
```

maximbaz:
Ready for your testing and a second round of reviews @erayd, @mcnesium 😉

src/background.js
erayd: > Could you please clarify? Based on how our password entry parsing looks now, I don't think the value can ever be null.

You're right; our key / value parsing disallows an empty value. I forgot that this was the case - and therefore, we don't need to detect an empty value here.

> Also, given that this function is only called with either ["login", "secret"] (which are guaranteed to always be present) or ["openid"] (which is only the case when "openid" was parsed, so again it's guaranteed to always be present), I don't really understand what this entire loop is trying to catch...

I have that PR for multi-stage login pending; once that is eventually merged then we'll be calling this with other combinations of options. So tests need to be _"does this set include these fields" _ rather than _"is this set exactly these fields"_.

Given that there isn't an empty-string problem, loose-type comparisons on the new L685 aren't a problem, but the explicit sets are.

src/inject.js
```diff
@@ -176,13 +200,11 @@
                     result.needPressEnter = true;
                 }
                 // We need to keep focus somewhere within the form, so that Enter hopefully submits the form.
-                var password = find(PASSWORD_FIELDS, loginForm);
-                if (password) {
```
erayd: Can we still preferentially focus the password field, if it's present and we filled it? That seems like the most reliable way of making sure we are inside the form.

src/background.js
```diff
@@ -298,49 +291,55 @@ async function fillFields(settings, login, fields) {
...
-    let allowNoSecret = false;
-    let filledFields = await dispatchFill(
-        settings,
-        fillRequest,
-        allFrames,
-        allowForeign,
-        allowNoSecret
+    let allowNoSecret = !fields.includes("secret");
+    let filledFields = [];
+    let importantFieldToFill = fields.includes("secret") ? "secret" : "openid";
```
erayd: I like this.

Can you clarify why this prefers 'secret'? I thought that 'openid' was supposed to be the preferred option. Or am I misunderstanding what this bit is for?

src/background.js
maximbaz: OK makes sense, I'll remove the check for now, add it back in your PR so that it is highlighted as "new code" and we remember to talk about it again in the context of multi-stage login 👍

src/inject.js
maximbaz: It is still the case, I only added OpenID fields. We loop over `[PASSWORD_FIELDS, USERNAME_FIELDS, OPENID_FIELDS]`, if a password field was found, that's where we put focus, if not found we try username field and place focus there, and again if not found we try to find OpenID field and put focus there.

Given that we submit either password+username or openid, in most cases we will focus password field or openid field.

src/background.js
maximbaz: This relies on the assumption that we never fill "secret" AND "openid" at the same time, so if `fields` contains "secret", it is guaranteed that it doesn't contain "openid". I can still inverse the logic, to highlight that "openid" is "more important".

But the purpose of this bit is just to correctly determine if we need to try to fill forms again in other frames, or we have already successfully filled the form. Suppose we are given `fields=["secret", "user"]`, that means that after filling a form we need to check if "secret" was filled - thus "secret" is the "important field to fill". But if we were given `fields=["openid"]`, we instead tell that a form was correctly filled when "openid" field was filled, therefore "openid" is the "important field to fill". Makes sense? 🙂

mcnesium:
This works perfectly now, thank you! :1st_place_medal:

PR #72, opened 2019-04-06
Support configuring badge colors

maximbaz:
To configure colors, you either define colors in `.browserpass.json` in the root of your password store (see README), or go to extension options and create a new password store.

Both background and text color can be optionally configured, leave any of them out and the default color will be applied.

Sorry no color picker, maybe later, maybe never.

image

Fixes https://github.com/browserpass/browserpass/issues/270

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,12 +166,14 @@ Using the `Custom store locations` setting in the browser extension options, you
 
 The list of available options:
 
-| Name                                                        | Description                                                  |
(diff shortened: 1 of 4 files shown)


maximbaz:
Maybe, but it needs to be a pre-defined set of good colors, not just anything random. But given that it's configurable via `.browserpass.json`, agree that it's low prio 👍

PR #80, opened 2019-04-09
Document OpenID

### LINKED ISSUE #79: Document support for OpenId
It would be cool to add a paragraph about OpenID support in the README.

maximbaz:
@erayd, @DamienCassou anything else you want to see mentioned, or this short summary is enough?

Fixes #79

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ In order to use Browserpass you must also install a [companion native messaging 
     -   [First steps in browser extension](#first-steps-in-browser-extension)
     -   [Available keyboard shortcuts](#available-keyboard-shortcuts)
     -   [Password matching and sorting](#password-matching-and-sorting)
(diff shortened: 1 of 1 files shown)


DamienCassou:
> Looks good :-). Do you think it's worth mentioning why we don't fill user / pass when openid: is present, or do you consider that unnecessary?

The current phrasing might be understood as a limitation of browserpass. I think a different phrasing might bring some clarity and answer your concern. Something like: When an `openid:` field is defined in a password entry and an openid form field is found in the current page, then Browserpass will fill the OpenID field in a form. In this case, Browserpass will not attempt to fill username and password even if they are present in the password entry because logging in with OpenID does not require them.

maximbaz:
Thanks guys, I added clarification 👍

PR #83, opened 2019-04-10
Switch to detecting outermost domain component

maximbaz:
@erayd need your opinion, why do we detect deepest available domain component instead of the outermost available one? We don't expect users to store "google.com/github.com.gpg", right? What was the use-case for this?

The reason I'm bringing this up is because someone reported to me a bug in person, that Browserpass does not correctly show the matching sites in popup.

For example, `~/.password-store/github.com/maxim.baz.gpg` will not show up on `github.com`, because according to our current parsing algorithm `maxim.baz` will be considered a domain name, not `github.com`.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -53,7 +53,7 @@ chrome.runtime.onInstalled.addListener(onExtensionInstalled);
 //----------------------------------- Function definitions ----------------------------------//
 
 /**
- * Get the deepest available domain component of a path
+ * Get the outermost available domain component of a path
  *
  * @since 3.0.0
  *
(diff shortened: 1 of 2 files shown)


erayd:
> We don't expect users to store "google.com/github.com.gpg", right?

Unfortunately, yes - we do. The specific use-case was storing credentials as (e.g.) `my-employer.com/gmail.com`. For example, someone working at Twitter might have an entry named `twitter.com/gmail.com`.

One of the reasons we added tldjs was to prevent problems with this approach - ensuring that only valid domains are considered as part of this algorithm removes most of the incorrect matches. It's still possible that the odd one might sneak through, but they should be the minority.

maximbaz:
Agree, we chatted offline and fixed the bug in a different way: #84

PR #84, opened 2019-04-10
Also check that the TLD actually exists

erayd:
Check the `tldExists` component of the TldJS object, to ensure that the TLD is actually valid, not just well-formed.

### DIFF
--- a/src/popup/popup.js
+++ b/src/popup/popup.js
@@ -44,7 +44,7 @@ function pathToDomain(path) {
             continue;
         }
         var t = TldJS.parse(parts[key]);
-        if (t.isValid && t.domain !== null) {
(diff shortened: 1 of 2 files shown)


src/popup/popup.js
```diff
@@ -44,7 +44,7 @@ function pathToDomain(path) {
             continue;
         }
         var t = TldJS.parse(parts[key]);
-        if (t.isValid && t.domain !== null) {
+        if (t.isValid && t.tldExists && t.domain !== null) {
```
maximbaz: For my future reference, we check `t.domain` because:

```
domain: null
hostname: "com"
isIp: false
isValid: true
publicSuffix: "com"
subdomain: null
tldExists: true
```

PR #85, opened 2019-04-10
Handle fields hidden with an opacity rule

maximbaz:
For example fixes auto-fill on app.wire.com when removing a device.

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -224,11 +224,11 @@
      * @return array List of search results
      */
     function queryAllVisible(parent, field, form) {
-        var result = [];
-        for (var i = 0; i < field.selectors.length; i++) {
-            var elems = parent.querySelectorAll(field.selectors[i]);
-            for (var j = 0; j < elems.length; j++) {
-                var elem = elems[j];
+        const result = [];
+        for (let i = 0; i < field.selectors.length; i++) {
+            let elems = parent.querySelectorAll(field.selectors[i]);
+            for (let j = 0; j < elems.length; j++) {
+                let elem = elems[j];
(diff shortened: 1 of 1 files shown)


src/inject.js
```diff
@@ -224,11 +224,11 @@
      * @return array List of search results
      */
     function queryAllVisible(parent, field, form) {
-        var result = [];
-        for (var i = 0; i < field.selectors.length; i++) {
-            var elems = parent.querySelectorAll(field.selectors[i]);
-            for (var j = 0; j < elems.length; j++) {
-                var elem = elems[j];
+        const result = [];
```
erayd: Why is this now declared as constant, when we are still modifying it on line 280?

src/inject.js
```diff
@@ -247,19 +247,35 @@
...
                     (rect.x > window.innerWidth || rect.y > window.innerHeight)
                 ) {
                     continue;
                 }
+                // Elem is hidden by its or or its parent's opacity rules
+                const OPACITY_LIMIT = 0.1;
+                let opacity = 1;
+                for (
+                    let testElem = elem;
+                    opacity >= OPACITY_LIMIT && testElem && testElem.nodeType === 1;
```
erayd: Please use the [appropriate constant](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType) here, not a magic number.

src/inject.js
maximbaz: `const` prevents you from assigning something entirely new (thus inadvertently overriding existing data), but it is allowed to modify internal structure of an array/object stored in a constant.

PR #86, opened 2019-04-10
[BLOCKED] Add manually crafted 16x16 icon

maximbaz:
Bye bye SVG, we have to switch back to PNG to support manually crafted 16x16 icon.

Not loading other formats because according to my experiments browsers scale 128px icon themselves very good, so we can save a bit of space.


maximbaz:
Nobody liked this icon anyway, and as I wrote in the issue, I think we should stop and say "good enough".

PR #87, opened 2019-04-11
Fix login settings parsing

maximbaz:
Will comment inline where the important part is, everything else is beautification while I was in that part of the file 🙂

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -778,7 +778,7 @@ async function parseFields(settings, login) {
         for (var key in login.fields) {
             if (
                 Array.isArray(login.fields[key]) &&
-                login.fields[key].indexOf(parts[0].toLowerCase()) >= 0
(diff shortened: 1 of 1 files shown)


src/background.js
```diff
@@ -789,7 +789,7 @@ async function parseFields(settings, login) {
         for (var key in login.settings) {
             if (
                 typeof login.settings[key].type !== "undefined" &&
-                login.settings[key].name.indexOf(parts[0].toLowerCase()) >= 0
+                login.settings[key].name === parts[0].toLowerCase()
```
maximbaz: This is the bug fix, contrary to `login.fields[key]`, `login.settings[key]` is not an array, so we must be comparing strings here directly.

This caused for example having a property `a: b` in a pass file to think that user wants to set `autoSubmit` to `false`, because `a` is included in the string `"autoSubmit"`!

PR #94, opened 2019-04-13
Improve error handling with native app comms for settings

### LINKED ISSUE #92: Options screen is replaced by error when password store directory doesn't exist
image

erayd:
Fixes #92.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -392,7 +392,7 @@ async function getFullSettings() {
     });
     var response = await hostAction(configureSettings, "configure");
     if (response.status != "ok") {
-        throw new Error(JSON.stringify(response)); // TODO handle host error
+        settings.hostError = response;
     }
     settings.version = response.version;
     if (Object.keys(settings.stores).length > 0) {
--- a/src/background.js
+++ b/src/background.js
@@ -411,7 +411,7 @@ async function getFullSettings() {
                 }
             }
         }
-    } else {
+    } else if (response.status == "ok") {
         // no user-configured stores, so use the default store
         settings.stores.default = {
             id: "default",
--- a/src/background.js
+++ b/src/background.js
@@ -593,7 +593,7 @@ async function handleMessage(settings, message, sendResponse) {
             } catch (e) {
                 sendResponse({
                     status: "error",
-                    message: "Unable to save settings" + e.toString()
+                    message: e.message
                 });
             }
             break;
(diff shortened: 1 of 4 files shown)


src/background.js
```diff
@@ -863,11 +863,6 @@ async function saveSettings(settings) {
     // 'default' is our reserved name for the default store
     delete settingsToSave.stores.default;
 
-    var response = await hostAction(settingsToSave, "configure");
-    if (response.status != "ok") {
-        throw new Error(JSON.stringify(response)); // TODO handle host error
-    }
```
maximbaz: I think we should keep this, it allows to quickly validate that users didn't make a typo, for example by writing path to a custom store, _before_ we actually save them in local storage.

It was more important before, because if you happen to save invalid options, options will refuse to open up next time, but even now I think it is better to not save invalid settings.

src/options/interface.js
```diff
@@ -76,6 +76,10 @@ function view(ctl, params) {
     if (typeof this.error !== "undefined") {
         nodes.push(m("div.error", this.error.message));
     }
+    if (this.settings.hasOwnProperty("hostError")) {
+        let hostError = this.settings.hostError;
+        nodes.push(m("div.error", hostError.params.message));
+    }
```
maximbaz: We need to do the same in popup, after this change when you rename your `~/.password-store` dir, popup will show no errors, but it should show the error from configure

src/options/options.js
```diff
@@ -42,6 +43,14 @@ async function saveSettings(settings) {
     if (response.status != "ok") {
         throw new Error(response.message);
     }
+
+    // reload settings
+    var response = await chrome.runtime.sendMessage({ action: "getSettings" });
+    if (response.status != "ok") {
+        throw new Error(response.message);
+    }
```
maximbaz: Was this added because you removed pre-save options validation, or this is needed for something else?

maximbaz:
I get a Type Error when saving an invalid custom password store:

image

src/options/options.js
erayd: This is needed to ensure that native host errors are resolved, and any new errors are displayed. Basically, it updates the interface settings object prior to redraw.

src/background.js
erayd: I missed that this is what it was doing - there was no comment on it, and when I looked at it I thought it was just fetching the settings and doing nothing with them. Will put it back.

maximbaz: Add comment as well 👍

erayd:
> I get a Type Error when saving an invalid custom password store

That should be fixed now. The pre-save check catches this.

PR #97, opened 2019-04-13
Fix loading store settings when "configure" failed

maximbaz:
Review with whitespace diff off: https://github.com/browserpass/browserpass-extension/pull/97/files?w=1

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -395,37 +395,41 @@ async function getFullSettings() {
         settings.hostError = response;
     }
     settings.version = response.version;
-    if (Object.keys(settings.stores).length > 0) {
-        // there are user-configured stores present
-        for (var storeId in settings.stores) {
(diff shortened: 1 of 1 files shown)


src/background.js
```diff
@@ -395,37 +395,41 @@ async function getFullSettings() {
...
+                    var storeSettings = settings.stores[storeId].settings;
+                    for (var settingKey in fileSettings) {
+                        if (!storeSettings.hasOwnProperty(settingKey)) {
+                            storeSettings[settingKey] = fileSettings[settingKey];
+                        }
                     }
                 }
             }
-        }
-    } else if (response.status == "ok") {
```
erayd: Why has this check been reversed? There's stuff in this block which requires a successful response.

erayd: Apparently it just looked this way reviewing the diff on my phone. It's fine.

PR #105, opened 2019-04-14
Add an option to disable the domain-count badge

erayd:
Disable the matching-domain-count badge over the toolbar icon.

Blocked by #106.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -87,7 +87,6 @@ function pathToDomain(path) {
 async function updateMatchingPasswordsCount(tabId) {
     try {
         const settings = await getFullSettings();
-        console.log(JSON.stringify(settings));
         if (settings.disableBadge === true) {
             return;
         }
--- a/src/background.js
+++ b/src/background.js
@@ -11,6 +11,7 @@ var appID = "com.github.browserpass.native";
 // default settings
 var defaultSettings = {
     autoSubmit: false,
+    disableBadge: false,
     gpgPath: null,
     stores: {},
     foreignFills: {},
@@ -86,6 +87,11 @@ function pathToDomain(path) {
 async function updateMatchingPasswordsCount(tabId) {
     try {
         const settings = await getFullSettings();
+        console.log(JSON.stringify(settings));
+        if (settings.disableBadge === true) {
+            return;
+        }
+
         var response = await hostAction(settings, "list");
         if (response.status != "ok") {
             throw new Error(JSON.stringify(response));
--- a/src/options/interface.js
+++ b/src/options/interface.js
@@ -72,7 +72,7 @@ function view(ctl, params) {
 
     nodes.push(m("h3", "Advanced settings"));
     nodes.push(
-        createCheckbox.call(this, "disableBadge", "Disable the matching-domains-count badge")
+        createCheckbox.call(this, "disableBadge", "Disable the badge counter")
     );
 
     if (typeof this.error !== "undefined") {
--- a/src/options/interface.js
+++ b/src/options/interface.js
@@ -45,7 +45,7 @@ function attach(element) {
  */
 function view(ctl, params) {
     var nodes = [];
-    nodes.push(m("h3", "Basic settings"));
+    nodes.push(m("h3", "Login settings"));
     nodes.push(createCheckbox.call(this, "autoSubmit", "Automatically submit forms after filling"));
     nodes.push(createInput.call(this, "username", "Default username", "john.smith"));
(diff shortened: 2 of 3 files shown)


src/options/interface.js
```diff
@@ -70,6 +70,11 @@ function view(ctl, params) {
         )
     );
 
+    nodes.push(m("h3", "Advanced settings"));
+    nodes.push(
+        createCheckbox.call(this, "disableBadge", "Disable the matching-domains-count badge")
+    );
```
maximbaz: CSS needs some adjustment, I think there's some margin-bottom under "add store", we should probably change it to margin-top on the Save button.

image

src/options/interface.js
```diff
@@ -70,6 +70,11 @@ function view(ctl, params) {
         )
     );
 
+    nodes.push(m("h3", "Advanced settings"));
+    nodes.push(
+        createCheckbox.call(this, "disableBadge", "Disable the matching-domains-count badge")
```
maximbaz: "Disable the badge counter" maybe? Because the counter might include recent entries from different domains as well.

src/background.js
```diff
@@ -86,6 +87,11 @@ function pathToDomain(path) {
 async function updateMatchingPasswordsCount(tabId) {
     try {
         const settings = await getFullSettings();
+        console.log(JSON.stringify(settings));
```
maximbaz: remember to remove

src/background.js
```diff
@@ -11,6 +11,7 @@ var appID = "com.github.browserpass.native";
 // default settings
 var defaultSettings = {
     autoSubmit: false,
+    disableBadge: false,
```
maximbaz: Add it to README

src/options/interface.js
erayd: Agreed. I want to deal with the whole options screen styling anyway (just opened #106 to keep track of this), so maybe let's hold off merging this until I've completed that task.

src/options/interface.js
erayd: I would consider a recent entry to be 'matching', but I see what you mean about the wording being ambiguous. Your wording is better, I'll change it.

src/options/interface.js
```diff
@@ -70,6 +70,11 @@ function view(ctl, params) {
         )
     );
 
+    nodes.push(m("h3", "Advanced settings"));
```
maximbaz: How about we replace "Basic settings" and "Advanced settings" with "Login settings" and "UI settings"? Theme selector might later go under the "UI settings" as well. And hopefully these group names would be more descriptive.

erayd: I like this idea - will do.

src/background.js
erayd: Thanks - well spotted :-).

src/background.js
```diff
@@ -86,6 +87,10 @@ function pathToDomain(path) {
 async function updateMatchingPasswordsCount(tabId) {
     try {
         const settings = await getFullSettings();
+        if (settings.disableBadge === true) {
+            return;
+        }
```
maximbaz: given that this setting can only be specified in browser extension, maybe instead of an early return here after requesting _full_ settings (which issues a request to the native app) let's update `chrome.tabs.onUpdated.addListener(....)` function, with something like:

```diff
// show number of matching passwords in a badge
-updateMatchingPasswordsCount(tabId);
+const settings = getLocalSettings();
+if (settings.disableBadge !== true) {
+    updateMatchingPasswordsCount(tabId);
+}
```

erayd: I was initially going to do that, until I realised that `updateMatchingPasswordsCount` is called from two different places - so I decided to make it a noop / early return.

Do you think the added maintenance complexity of tracking multiple calls to `updateMatchingPasswordsCount` is worth it for the performance gained by not loading full settings?

maximbaz: Not worth the complexity, agree 👍

erayd: OK.

src/options/interface.js
```diff
@@ -70,6 +70,9 @@ function view(ctl, params) {
         )
     );
 
+    nodes.push(m("h3", "UI settings"));
+    nodes.push(createCheckbox.call(this, "disableBadge", "Disable the badge counter"));
```
maximbaz: for after options screen is redesigned: it might look better if UI settings section is placed right below Login settings, or it might not - just something to consider 🙂

erayd: Good idea. We can have a look at it both ways and see once #106 is done.

PR #107, opened 2019-04-14
Assume current domain is valid, even if the TLD is not

### LINKED ISSUE #99: Treat the current domain in the address bar as valid
The current domain in the address bar should be considered valid when filtering entries, even if it uses a nonexistent TLD.

erayd:
When checking path elements for domain-ness, assume that the current tab hostname is a valid domain, even if it doesn't use a legitimate TLD.

Fixes #99.

### DIFF
--- a/src/popup/popup.js
+++ b/src/popup/popup.js
@@ -45,7 +45,12 @@ function pathToDomain(path, currentHost) {
             continue;
         }
         var t = TldJS.parse(parts[key]);
-        if (t.isValid && ((t.tldExists && t.domain !== null) || t.hostname === currentHost)) {
+        if (
+            t.isValid &&
+            ((t.tldExists && t.domain !== null) ||
+                t.hostname === currentHost ||
+                currentHost.endsWith(`.${t.hostname}`))
+        ) {
             return t.hostname;
         }
     }
--- a/src/popup/popup.js
+++ b/src/popup/popup.js
@@ -34,17 +34,18 @@ function handleError(error, type = "error") {
  *
  * @since 3.0.0
  *
- * @param string path Path to parse
+ * @param string path        Path to parse
+ * @param string currentHost Current hostname for the active tab
  * @return string|null Extracted domain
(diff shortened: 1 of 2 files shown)


src/popup/popup.js
```diff
@@ -34,17 +34,18 @@ function handleError(error, type = "error") {
...
-function pathToDomain(path) {
+function pathToDomain(path, currentHost) {
     var parts = path.split(/\//).reverse();
     for (var key in parts) {
         if (parts[key].indexOf("@") >= 0) {
             continue;
         }
         var t = TldJS.parse(parts[key]);
-        if (t.isValid && t.tldExists && t.domain !== null) {
+        if (t.isValid && ((t.tldExists && t.domain !== null) || t.hostname === currentHost)) {
```
maximbaz: To support the scenario of when `currentHost` is `login.pi.hole` and the entry is called `pi.hole.gpg`, I think we have to repeat the logic from line 98 here:

```
|| (currentHost == t.hostname || currentHost.endsWith("." + t.hostname))
```

src/popup/popup.js
```diff
@@ -34,17 +34,18 @@ function handleError(error, type = "error") {
  *
  * @since 3.0.0
  *
- * @param string path Path to parse
+ * @param string path        Path to parse
+ * @param string currentHost Current hostname for the active tab
  * @return string|null Extracted domain
  */
-function pathToDomain(path) {
+function pathToDomain(path, currentHost) {
```
maximbaz: because we haven't implemented some form of `helper.js` yet, remember to update the same function in `background.js` (sorry 😛)

src/popup/popup.js
erayd: Ooh, good catch! Will do.

src/popup/popup.js
erayd: This got cloned? Will do. `helper.js` suddenly rose in importance...

PR #108, opened 2019-04-14
Implement usage log, document usage metadata

### LINKED ISSUE #27: Keep audit data (which password was used where and when)
I learned today that I had a malicious plugin installed for a specific known to me period of time, I wanted to find out which websites could have been compromised. Then I remembered about a wonderful `recent` object in localStorage of browserpass. 

My ideas was: 
- Download the entire object
- Keep only entries whose `when` is newer than the start time of the attack.
- For each remaining entry, rotate password

Unfortunately, this idea didn't work because we extensively hash the key 🙂 

https://github.com/browserpass/browserpass-extension/blob/98b6201f58fc7ed75ebdd7f0a60f8b831fb93a50/src/popup/popup.js#L159

I was wondering if hashing gives us any good benefits, and if no, can we not hash the key?

Alternatively, how about just adding a couple more fields in the object, for audit purposes? Now we have `count` and `when`, we could extend `when` to be a list, and add `host`, `store.id` and `login` in there.

maximbaz:
Fixes #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ Browserpass allows configuring certain settings in different places places using
 
 Browserpass keeps metadata of recently used credentials in local storage and Indexed DB of the background page. This is first and foremost internal data to make Browserpass function properly, used for example to implement the [Password matching and sorting](#password-matching-and-sorting) algorithm, but nevertheless you might find it useful to explore using your browser's devtools. For example, if you are considering to rotate all passwords that you used in the past month (e.g. if you just found out that you had a malicious app installed for several weeks), you can retrieve such list from Indexed DB quite easily (open an issue if you need help).
 
-For details on how we treat this and any other data, consult [Security](#security) and [Privacy](#privacy) sections.
+For details on how we treat your data and how to remove it, consult [Security](#security) and [Privacy](#privacy) sections.
 
 ## Security
 
@@ -236,7 +236,9 @@ Browserpass aims to protect your passwords and computer from malicious or fraudu
 
 Browserpass does not send any telemetry data. All metadata that is collected in order for the extension to function correctly is stored _only_ in local storage, and never leaves your browser.
 
-This data is not synchronized by Google between your computers, and upon removing Browserpass extension all the data will be automatically purged by your browser.
+This data is not synchronized between your computers, and upon removing Browserpass extension all the data will be automatically purged by your browser.
+
+In order to remove all metadata, use "Clear all data" button in the extension options page or do it using your browser's devtools.
 
 ## Requested permissions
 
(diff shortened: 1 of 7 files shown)


src/background.js
```diff
@@ -177,6 +178,14 @@ function saveRecent(settings, login, remove = false) {
     if (!login.inCurrentDomain && login.recent.count === 1) {
         updateMatchingPasswordsCount(settings.tab.id);
     }
+
+    // save to usage log
+    const db = await idb.openDB("browserpass", 1, {
```
erayd: Please define a version constant, rather than using a magic number. As-is, it's wasn't obvious what this meant until I looked up the API documentation for `idb::openDB()`.

README.md
```diff
@@ -229,6 +236,8 @@ Browserpass aims to protect your passwords and computer from malicious or fraudu
 
 Browserpass does not send any telemetry data. All metadata that is collected in order for the extension to function correctly is stored _only_ in local storage, and never leaves your browser.
 
+This data is not synchronized by Google between your computers, and upon removing Browserpass extension all the data will be automatically purged by your browser.
```
erayd: It's not synchronised *by anyone* between computers. Not just Google.

erayd:
After further discussion we have agreed that we won't clutter the options with this, because the concern is largely theoretical rather than practical, but we will add a 'clear all data' button so that worried users are able to easily nuke it at will.

src/background.js
```diff
@@ -862,6 +874,26 @@ async function receiveMessage(message, sender, sendResponse) {
...
+ *
+ * @since 3.0.10
+ *
+ * @return void
+ */
+async function clearData() {
+    // clear local storage
+    localStorage.removeItem("foreignFills");
+    Object.keys(localStorage).forEach(key => {
+        if (key.startsWith("recent")) {
```
erayd: Can we change this to `recent:`, and then also explicitly remove the `recent` key? If we add any other settings named recent* in the future, we may not remember this, but the button would still nuke it.

src/options/interface.js
```diff
@@ -97,6 +99,24 @@ function view(ctl, params) {
...
+                    try {
+                        await this.clearData();
+                        this.error = undefined;
+                    } catch (e) {
+                        this.error = e;
+                    }
+                    m.redraw();
+                }
+            },
+            "Clear all data"
```
erayd: Perhaps "Clear usage data"? "All" implies that it nukes all the settings, which isn't the case.

README.md
```diff
@@ -229,6 +236,10 @@ Browserpass aims to protect your passwords and computer from malicious or fraudu
 
 Browserpass does not send any telemetry data. All metadata that is collected in order for the extension to function correctly is stored _only_ in local storage, and never leaves your browser.
 
+This data is not synchronized between your computers, and upon removing Browserpass extension all the data will be automatically purged by your browser.
+
+In order to remove all metadata, use "Clear usage data" button in the extension options page or do it using your browser's devtools.
```
erayd: ...use **the** "Clear usage data" button...

PR #118, opened 2019-04-15
Pass OTP data across to OTP extension for further action

erayd:


src/background.js
```diff
@@ -851,6 +868,23 @@ async function parseFields(settings, login) {
...
         }
     }
+
+    // trigger otp extension
+    if (login.fields.hasOwnProperty("otp")) {
+        for (let key in otpID) {
+            chrome.runtime
+                .sendMessage(otpID[key], {
+                    otp: login.fields.otp,
+                    host: new URL(settings.tab.url).hostname,
```
maximbaz: this can be simply `host: settings.host` I think

Issue #98, opened 2019-04-13
Add light theme


adtac:
### General information

-   Operating system + version: Arch Linux
-   Browser + version: Firefox Nightly (2019-04-13)
-   Information about the host app:
    -   How did you install it? `browserpass` from [community]
    -   If installed an official release, put a version (`$ browserpass --version`): 3.0.6
-   Information about the browser extension:
    -   How did you install it? Firefox Add-ons store
    -   Browserpass extension version as reported by your browser: 3.0.9

---

Browserpass v2 used to have a light theme and it was great. But it looks like v3 only supports a dark theme. I know that dark themes are more popular with developers (I, too, used to use dark themes), but a small minority of us prefer light themes. FWIW, dark text on light backgrounds is supposed to be more legible and readable to human eyes if the room you're in is well lit (which is the case for me).

As a worst case scenario, allow adding custom override CSS like available in the Tab Center Redux extension. But I'd much rather prefer a maintained light theme because div IDs and classes may change across releases, making custom CSS difficult to maintain.

maximbaz:
Initially I had my reservations about supporting both themes (just because of the increased maintenance cost), but I believe it's a valid request and if we implement it right so that increased cost is minimal, I won't oppose a PR.

I propose to move everything color-related out of `popup.less` into `popup-dark.less`, and create a new `popup-light.less` file. Hopefully most of the styles will remain in a shared file, and the theme-related files will be minimal. Then add a class `theme-dark` or `theme-light` on `<body>` tag based on the current settings (keep dark theme a default), and in `popup-*.less` wrap everything in `.theme-dark { ... }` or `.theme-light { ... }` accordingly.

I don't think I'll be working on a light theme myself, so a PR is very welcome 🙂 

---------

If you would like to see a light theme, please upvote the issue, once someone proposes a PR I'll provide you with a test build and ping you for feedback.

--------

P.S. Since you are on Arch, I'll use this opportunity to recommend you to replace the addon from Web store with `browserpass-firefox` from `[community]`. I've just realized that this repo doesn't even mention this package in README, will fix in a moment.

adtac:
Thanks, I've switched to the AUR version now.

And I'll try to submit a PR later this week. I can't maintain it, but I'm willing to do the initial work required.

erayd:
@adtac Thank you - will keep an eye out for your PR.

Once it's merged, we will maintain it as part of the main codebase.

maximbaz:
[...]

Issue #61, opened 2016-12-31
Add ability to generate/add passwords


emersion:
It would be nice to be able to generate and add passwords directly from the extension.

adtac:
I just want to add: can the implementation make the password generation method configurable? I wrote a small patch to add [XKCD style passwords](https://xkcd.com/936/) (see [mailing list](https://lists.zx2c4.com/pipermail/password-store/2018-October/003433.html), and I'd like to have the option to generate these. I think we'll need such an option in the settings any way to support flags like `-n` to generate passwords without symbols? I propose three options in the add-on settings:

 - Default password
 - No symbols (`-n`)
 - XKCD/Diceware passphrases

maximbaz:
Length and characters set will definitely be configurable, but probably not XKCD passphrases, not convinced it is worth the effort in the context of a password manager :) Seems like some folks from that thread think similarly...

adtac:
I'm happy with that. As long as something is configurable, it'll be pretty easy for me to maintain a fork with the extra code (which is what I'm doing with the shell script right now).

maximbaz:
Hi guys!

It's been a long time since I wanted to implement editing feature, but it seems I won't be able to finish it in timely fashion 😞 

I would therefore like to ask if anyone is interested to pick it up, finish the work and contribute PRs?

The necessary changes would include:

- Updating the popup UI
- Extending browserpass-native with new methods to create/edit/delete password entries

Because it's a significant change, I believe it would make sense to split the work in several PRs, so that it's easier to review and address feedback, to make sure we are moving in the right direction and not implementing unnecessary functionality.

For example, I would start with creating an interface, focusing on bigger picture and simply disabling buttons like "Save", "Delete entry" and "Generate password", just to focus on the UI first. Filling the edit form with existing credentials would not require extending `browserpass-native`, it already contains all the necessary APIs. Once that is done, saving the edits, adding new entries, deleting existing entries, extending options screen with e.g. password generation configurations, can all come as separate PRs.

I have started the design implementation some time ago in [this branch](https://github.com/maximbaz/browserpass-extension/commits/add-edit), I think some parts of the code could still be reused.

Here's how I see the interface looking:

![demo](https://user-images.githubusercontent.com/1177900/56982993-d8e02700-6b82-11e9-9257-ca3fcab8d1ee.gif)

[...]

PR #120, opened 2019-04-16
Add light theme

adtac:
Here's what it looks like:

light

Closes https://github.com/browserpass/browserpass-extension/issues/98

### DIFF
--- a/src/popup/colors.less
+++ b/src/popup/colors.less
@@ -1,0 +1,76 @@
+.colors(
+    @bg-color,
+    @default-bg-color,
+    @hover-bg-color,
+    @text-color,
+    @error-text-color,
+    @dim-text-color,
+    @badge-color,
+    @input-bg-color,
+    @active-input-bg-color,
+    @input-text-color,
+    @hint-bg-color,
+    @hint-color) {
+    @match-text-color: @hint-bg-color;
+
+    html,
+    body {
+        background-color: @bg-color;
+        color: @text-color;
+    }
+
+    .badge {
+        background-color: @badge-color;
+    }
+
+    .part.error {
+        color: @error-text-color;
+    }
+
+    .part.search {
+        background-color: @input-bg-color;
+    }
+
+    .part.search:focus-within {
+        background-color: @active-input-bg-color;
+    }
+
+    .part.search > .hint {
+        background-color: @hint-bg-color;
+        color: @hint-color;
+    }
+
+    .part.search > input[type="text"] {
+        background-color: transparent;
+        color: @input-text-color;
+    }
+
+    .part.search > input[type="text"]::placeholder {
+        color: @dim-text-color;
+    }
+
+    .logins:not(:hover):not(:focus-within) .part.login:first-child > .name {
+        background-color: @default-bg-color;
+    }
+
+    .part.login > .name:hover,
+    .part.login > .name:focus,
+    .part.login > .action:hover,
+    .part.login > .action:focus,
+    .part.login:focus > .name {
+        background-color: @hover-bg-color;
+    }
+
+    .part.login em {
+        color: @match-text-color;
+    }
+
+    .updates {
+        border-top: 1px solid @hover-bg-color;
+
+        span,
+        a {
+            color: @error-text-color;
+        }
+    }
+}
--- a/src/popup/colors.less
+++ b/src/popup/colors.less
@@ -10,9 +10,9 @@
     @active-input-bg-color,
     @input-text-color,
     @hint-bg-color,
-    @hint-color) {
-    @match-text-color: @hint-bg-color;
-
+    @hint-color,
+    @match-text-bg-color,
+    @match-text-color) {
     html,
     body {
         background-color: @bg-color;
@@ -62,6 +62,7 @@
     }
 
     .part.login em {
+        background-color: @match-text-bg-color;
         color: @match-text-color;
     }
 
--- a/src/popup/popup.js
+++ b/src/popup/popup.js
@@ -43,6 +43,10 @@ async function run() {
         }
         var settings = response.settings;
 
+        var root = document.getElementsByTagName("html")[0];
+        root.classList.remove("colors-dark");
+        root.classList.add(`colors-${settings.theme}`);
+
         if (settings.hasOwnProperty("hostError")) {
             throw new Error(settings.hostError.params.message);
         }
--- a/src/Makefile
+++ b/src/Makefile
@@ -20,7 +20,7 @@ css/popup.dist.css: $(LESSC) popup/*.less
 	[ -d css ] || mkdir -p css
 	$(LESSC) popup/popup.less css/popup.dist.css
 
-css/options.dist.css: $(LESSC) options/options.less
+css/options.dist.css: $(LESSC) options/*.less
 	[ -d css ] || mkdir -p css
 	$(LESSC) options/options.less css/options.dist.css
 
--- a/src/Makefile
+++ b/src/Makefile
@@ -16,7 +16,7 @@ deps:
 prettier: $(PRETTIER) $(PRETTIER_FILES)
 	$(PRETTIER) --write $(PRETTIER_FILES)
 
-css/popup.dist.css: $(LESSC) popup/popup.less
+css/popup.dist.css: $(LESSC) popup/*.less
 	[ -d css ] || mkdir -p css
 	$(LESSC) popup/popup.less css/popup.dist.css
 
--- a/src/background.js
+++ b/src/background.js
@@ -23,7 +23,8 @@ var defaultSettings = {
     gpgPath: null,
     stores: {},
     foreignFills: {},
-    username: null
+    username: null,
+    theme: "dark"
 };
 
 var authListeners = {};
--- a/src/options/interface.js
+++ b/src/options/interface.js
@@ -59,7 +59,12 @@ function view(ctl, params) {
     nodes.push(createInput.call(this, "gpgPath", "Custom gpg binary", "/path/to/gpg"));
 
     nodes.push(m("h3", "Theme"));
-    nodes.push(createDropdown.call(this, "theme", [m("option", { value: "dark" }, "Dark")]));
+    nodes.push(
+        createDropdown.call(this, "theme", [
+            m("option", { value: "dark" }, "Dark"),
+            m("option", { value: "light" }, "Light")
+        ])
+    );
 
     nodes.push(m("h3", "Custom store locations"));
     nodes.push(
--- a/src/options/interface.js
+++ b/src/options/interface.js
@@ -163,7 +163,7 @@ function createInput(key, title, placeholder) {
 /**
  * Generates vnode for a dropdown setting
  *
- * @since 3.0.13
+ * @since 3.3.1
  *
  * @param string key     Settings key
  * @param array options  Array of objects with value and text fields
--- a/src/options/interface.js
+++ b/src/options/interface.js
@@ -58,6 +58,9 @@ function view(ctl, params) {
     nodes.push(createInput.call(this, "username", "Default username", "john.smith"));
     nodes.push(createInput.call(this, "gpgPath", "Custom gpg binary", "/path/to/gpg"));
 
+    nodes.push(m("h3", "Theme"));
+    nodes.push(createDropdown.call(this, "theme", [m("option", { value: "dark" }, "Dark")]));
+
     nodes.push(m("h3", "Custom store locations"));
     nodes.push(
         m("div", { class: "notice" }, "(this overrides default store and $PASSWORD_STORE_DIR)")
@@ -150,6 +153,29 @@ function createInput(key, title, placeholder) {
             })
         ])
     ]);
+}
+
+/**
+ * Generates vnode for a dropdown setting
+ *
+ * @since 3.0.13
+ *
+ * @param string key     Settings key
+ * @param array options  Array of objects with value and text fields
+ * @return Vnode
+ */
+function createDropdown(key, options) {
+    return m(
+        "select",
+        {
+            value: this.settings[key],
+            onchange: e => {
+                this.settings[key] = e.target.value;
+                this.saveEnabled = true;
+            }
+        },
+        options
+    );
 }
 
 /**
--- a/src/popup/colors-dark.less
+++ b/src/popup/colors-dark.less
@@ -13,6 +13,8 @@
         @active-input-bg-color: #4f4f4f,
         @input-text-color: #eee,
         @hint-bg-color: #d79921,
-        @hint-color: #363636
+        @hint-color: #363636,
+        @match-text-bg-color: transparent,
+        @match-text-color: #d79921
     );
 }
--- a/src/popup/colors-dark.less
+++ b/src/popup/colors-dark.less
@@ -1,0 +1,18 @@
+@import "colors.less";
+
+.colors-dark {
+    .colors(
+        @bg-color: #414141,
+        @default-bg-color: #393939,
+        @hover-bg-color: #363636,
+        @text-color: #c4c4c4,
+        @error-text-color: #f00,
+        @dim-text-color: #a0a0a0,
+        @badge-color: #626262,
+        @input-bg-color: #4a4a4a,
+        @active-input-bg-color: #4f4f4f,
+        @input-text-color: #eee,
+        @hint-bg-color: #d79921,
+        @hint-color: #363636
+    );
+}
--- a/src/popup/colors-light.less
+++ b/src/popup/colors-light.less
@@ -13,7 +13,9 @@
         @active-input-bg-color: #e8ebee,
         @input-text-color: #343a40,
         @hint-bg-color: #1c7ed6,
-        @hint-color: #e7f5ff
+        @hint-color: #e7f5ff,
+        @match-text-bg-color: #cfecff,
+        @match-text-color: #1873ea
     );
 
     .part.login .name .line1 .recent,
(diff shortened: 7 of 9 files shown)


adtac:
I'm not very good with frontend stuff, so please feel free to point out anything stupid lol

adtac:
@erayd thanks for the review. Previously, match text colour was coupled with the hint background colour for some reason. I've decoupled the two now, so it should be much better. Here's what the popup looks like now:

Untitled

adtac:
(Made the badge a bit less dark to improve text contrast.)

erayd:
Definitely improving - thanks for your work on this :+1: .

Can you add some colour (or better contrast) back to the domain badge in the search field? It's fading into the background a bit:

image

## Other Stuff
There is no noticable difference in the search field background colour based on whether it's focused or not. Compare with the dark theme to see what I mean - the dark theme gets this bit right.

When moving focus from the search field to the first item (e.g. by pressing the down-arrow key), the first item gets much lighter, and gives the impression that it's just been deselected. Changing the colour or shade is fine, but it needs to indicate to the user that this element has *gained* focus, not lost it.

erayd:
Here's another screenshot showing what it looks like with multiple custom stores configured:
scr-20190416-165402

adtac:
@erayd updated. I was a bit hesitant to add non-grey colours to the mix as I didn't want it to be jarring, but I suppose it's alright to make the badge stand out a bit more.

Untitled

Hover is now darker than active so that it's more intuitive. The search field gets subtly lighter when it goes out of focus as well.

erayd:
Getting there. Now the domain badge has very low contrast with the 'remove-domain-badge' button though.

adtac:
Heh, you have a very keen eye, I missed that. Updated to a safer yellow :)

src/popup/colors.less
```diff
@@ -0,0 +1,78 @@
...
+        background-color: @default-bg-color;
+    }
+
+    .part.login > .name:hover,
+    .part.login > .name:focus,
+    .part.login > .action:hover,
+    .part.login > .action:focus,
+    .part.login:focus > .name {
+        background-color: @hover-bg-color;
+        outline: none;
```
maximbaz: I think outline needs to be declared in `popup.less`, not here

src/popup/popup.js
```diff
@@ -117,6 +117,10 @@ async function run() {
         }
         var popup = new Interface(settings, logins);
         popup.attach(document.body);
+
+        document
+            .getElementsByTagName("html")[0]
+            .setAttribute("class", "colors-" + settings["theme"]);
```
maximbaz: Use ``.classList.add(`colors-${settings.theme}`)``, so that we don't accidentally override an existing class on `<html>` tag (now or in future).

maximbaz: In addition, please move this snippet to an earlier stage in the function, right after we retrieve settings. Right now I see a visible change of colors, when I open the popup, at first it is something white, and then in a split second it changes to dark theme.

Managed to catch a screenshot of it:

image

I confirm that this fixes the issue for my particular case, but to be on a safe side, can we perhaps even add `class="colors-dark"` directly in `popup.html`, to keep the theme by default dark even while we are loading settings? And this snippet would need to be adjusted to something like:

```
const root = document.getElementsByTagName("html")[0];
root.classList.remove("colors-light", "colors-dark");
root.classList.add("colors-${settings.theme}");
```

maximbaz:
Finally got to try this! Here's some of my feedback:

1. I think we should be consistent in the way we highlight filter matches, if it is a different _text color_ in dark theme, let's use text color in light theme too, not _background color_:

    image

    image

1. The "Copy" buttons look like they are disabled, perhaps because the icon color is too light on a light background. Do you guys have the same impression?

1. The "Copy" buttons and the "recency" badge (and maybe even the password store badge) look particularly hard to distinguish when the mouse is hovering the row:

    image

    image

    image

    image

1. I agree that the domain filter should probably be light blue instead of red or yellow, a couple of suggestions just to give an impression, what do you think?

    background-color: #aad2f5;
    color: #343a40;

    image

    background-color: #39a3ff;
    color: white;

    image

maximbaz:
/cc a couple of people who I recall to be passionate about design and everyone who upvoted #98 so far, just to get a couple of extra opinions: @DamienCassou, @532910, @JanMa, @jakubtopic, @Tharre.

If anyone can't compile the branch to test, I'm willing to provide test builds, just say so.

I'd very much prefer to spend an extra time and get this right on the first try, so please, early feedback is very welcome.

Tharre:
The buttons look grayed out/disabled indeed.

The darker blue looks the nicest to me, but you might want to tone down the red if you go with it.

erayd:
> I think we should be consistent in the way we highlight filter matches, if it is a different *text color* in dark theme, let's use text color in light theme too, not *background color*.

I disagree with this point - I actually think the way he's done the highlighting is spot on. Yes, it's different, but I don't see any issue with different themes being different, and the downside of a light pallette is it's harder to get good contrast with a text-colour change, the way the dark theme does it.

adtac:
>I think we should be consistent in the way we highlight filter matches, if it is a different text color in dark theme, let's use text color in light theme too, not background color:

It's currently both a different text colour *and* a different background colour. And as @erayd said, I personally think a small background contrast adds a lot to the visibility of the search matches. In fact, I wanted to update the dark theme to support this as well (I can still do this if others like the light theme's way of doing search matches better).

>The "Copy" buttons look like they are disabled, perhaps because the icon color is too light on a light background. Do you guys have the same impression?

>The "Copy" buttons and the "recency" badge (and maybe even the password store badge) look particularly hard to distinguish when the mouse is hovering the row

Agreed, I've updated this. It was a bit tricky because the classes use a SVG background with a hard-coded fill colour. I didn't want to create a separate SVG for the light theme as that's just an unnecessary maintenance overhead. As a result, the light theme is using CSS invert. While this is a perfectly valid solution for light/dark themes, implementing other themes will likely require hue rotate and other shenanigans.

I've also updated the badge to a dark blue. I agree with @Tharre that it looks nicer.

>I confirm that this fixes the issue for my particular case, but to be on a safe side, can we perhaps even add class="colors-dark" directly in popup.html, to keep the theme by default dark even while we are loading settings? And this snippet would need to be adjusted to something like

If the dark theme is the default, I agree. I've made the change you suggested, but I'd also like to start a discussion on what the default theme itself should be. I think there's a case to be made that the default theme should be light for two reasons:

 - Most browsers come with a default light theme, so it's a bit jarring to have an extension not follow the standard.
 - v2 used to have a light theme, so sticking to a light theme would make the upgrade easier from a muscle memory perspective.

JanMa:
So far I like the light theme a lot. Great work @adtac :-)
Maybe you could think about using the same yellow colours for highlighting as in the dark theme but this is just personal preference. I would also favour the light theme to be the default one, because light themes are also the default in Chrome and Firefox.

DamienCassou:
I have used the light theme the whole day and I like it. Thank you for your work.

maximbaz:
Getting much better!

1. I couldn't leave without trying to highlight with text color alone (`#00bcff`), IMHO looks quite alright, is it worse in your opinion than the highlight with background color?

    image


1. If you still think the background highlight is much better, could it be made so that the blue background doesn't "touch" the store badge? In the picture below, ideally the blue background would stop maybe just a bit above the top of the "h" letter, but I feel there should be some space left between blue background and the store badge.

    image

1. Would you agree that there's not enough contrast in this case, between blue background and the background of this entry? It doesn't look as bad on the enlarged screenshot, better try this in your browser:

    image

1. When an entry is hovered, it feels the store badge also has small contrast (again doesn't look that bad on the enlarged screenshot, but try in your browser):

    image

erayd:
@maximbaz I still think the background highlight is better than the change in text color - the  text in your example fades into the background somewhat. I think the background highlight colour is fine as-is, and the contrast is sufficient. Remember that it contrasts with the text too, not just the background.

I agree that there needs to be space between the highlight and the badge. And perhaps the badge colour could change a little too improve contrast in your point (4).

adtac:
> 1. I couldn't leave without trying to highlight with text color alone (`#00bcff`), IMHO looks quite alright, is it worse in your opinion than the highlight with background color?

Not at all. I still think the light background makes it stand out a bit more. However, if you feel strongly about this, please don't hesitate to ask me to implement it without the background; it's your project, after all :) I don't mind either, but I prefer the background (and apparently, so do some others).

> 2. If you still think the background highlight is much better, could it be made so that the blue background doesn't "touch" the store badge? In the picture below, ideally the blue background would stop maybe just a bit above the top of the "h" letter, but I feel there should be some space left between blue background and the store badge.

Added a 2px margin between the lines. Does not affect the height of each entry.

> 3. Would you agree that there's not enough contrast in this case, between blue background and the background of this entry? It doesn't look as bad on the enlarged screenshot, better try this in your browser:

True, made it a bit darker.

> 4. When an entry is hovered, it feels the store badge also has small contrast (again doesn't look that bad on the enlarged screenshot, but try in your browser):

Fixed as well.

> 5. I'd recommend `invert(85%)` instead of `100%`, makes the icons look not as bold:

Done (although this makes the hover background logic quite tricky as we now need to use a colour such that `invert(85%)` is the same as `@hover-bg-color`, for which only an approximate solution exists). We could change the SVG's colour slightly so that `invert(100%)` isn't so contrasting, but this is a less "elegant" solution IMO.

> 6. The contrast between a focused search box and a pre-selected first line might also need additional improvement:

Fixed.

> 7. Is the "search icon" on purpose left without the "invert"? On the one hand it's not a button, on the other hand in the dark theme it has the same color as the buttons, and some might say it's almost invisible due to the low contrast with the background of the focused search input box:

This is intentional. Due to `invert` and the way the search button's CSS is written, this is impossible because applying a filter would make the search background dark.

The most elegant solution (for points 5, 7) would be to inline the SVGs in the LESS and use a variable-based fill colour inside the SVG string. Perhaps this could be done in a future improvement (unfortunately, I don't have the time to do this).

BTW, please feel free to make changes as you see fit once you merge this. I think that'd be a much more productive way than having back-and-forth replies about subtle colour changes :)

erayd:
> Added a 2px margin between the lines. Does not affect the height of each entry.
The entry height is the same (53px), but it's stuffed up the vertical height somehow. The light theme now pushes the sixth entry off the bottom:

image
image

Ideally, the bottom of the popup would be flush with the bottom of the sixth entry. I'm not sure why the bottom of the popup is collapsing this way, but it needs resolving. The dark theme also exhibits this a little, but only by a very small amount.

> Done (although this makes the hover background logic quite tricky as we now need to use a colour such that invert(85%) is the same as @hover-bg-color, for which only an approximate solution exists).
Can you explain why this complexity is needed? What are you trying to achieve by having them be the same? From a visual perspective what you've done here looks fine.

> This is intentional. Due to invert and the way the search button's CSS is written, this is impossible because applying a filter would make the search background dark.
I'm actually OK with this as it is. It makes sense for the search button to be pale, because it isn't a button, it's just a hint. I don't think we need to change it.

> BTW, please feel free to make changes as you see fit once you merge this. I think that'd be a much more productive way than having back-and-forth replies about subtle colour changes :)
I disagree. Much of the point of the PR process is to get things right the first time, and merge it when it's ready. There's no deadline that we have to merge this feature by, so we can take our time here and get it right. You've done a great job here - now we're just at the nitpicking stage trying to get it exactly right.

maximbaz:
Awesome work, I'll give it another go later tonight 😉 

> The entry height is the same (53px), but it's stuffed up the overall vertical height of the popup somehow. The light theme now pushes the sixth entry off the bottom

I actually don't reproduce this, even if I set custom zoom in Chromium settings.

> > This is intentional. Due to invert and the way the search button's CSS is written, this is impossible because applying a filter would make the search background dark.
> 
> I'm actually OK with this as it is. It makes sense for the search button to be pale, because it isn't a button, it's just a hint. I don't think we need to change it.

Me too, just wanted to point this out to make sure it wasn't omitted 🙂

erayd:
> I actually don't reproduce this, even if I set custom zoom in Chromium settings.

Then let's ignore it; it's likely something funky with my zoom / setup. Chrome certainly has a number of issues in that area.

maximbaz:
We can ignore it for now, but it would be a pity to degrade your experience, even though you run a custom setup. Let's return to this in the end, maybe there are some other ways to implement the margin that would not cause misalignment for you...

maximbaz:
With the current colors I'm satisfied with everything except the default password store badge color, it has a bit of "violet" flavor that imho doesn't fit with the rest of the popup.

Status quo:

image

Let's explore other colors?

My proposals are light blue `#a9caec`:

image

Or dark grey `#b2b2b2`:

image

Opinions, other suggestions?

maximbaz:
I've been using the light theme for a few hours, all in all it was quite a good experience, besides the password store badge color there was one other particular scenario that threw me completely off: when I had to press <kbd>Backspace</kbd> and search manually for credentials that don't belong to the current domain, I was genuinely distracted by the highlight of matches, I think they stand out too much than necessary.

Let's take this fake example: I want to find a password "github.com" in "personal" folder, so I know that typing "pegit" would be enough to bring it to the top - and it did, but all these blue spots made me feel distracted, I felt like I am asked to pay extra attention to them all:

image

I honestly didn't plan to bring up the topic of highlighting using background color again, so instead of forcing you into changing to use text color alone, let me just ask you this: could you please try this scenario for yourself and tell me if you also feel that it is distracting when the number of blue spots is big?

532910:
image

The one of a good design principles is to use as few colors/fonts as possible.

Why not to use #027b83 (yellow) instead of #d79921 (dark green)?
And same for light theme and blue color.

maximbaz:
Green looks worse on the dark grey background, and I'm so very fond of the combination of grey and orange, I don't want to change the dark theme 🙂

erayd:
> Let's explore other colors?
> My proposals are light blue `#a9caec`

Good call. My preference is for the blue you've proposed.

> Could you please try this scenario for yourself and tell me if you also feel that it is distracting when the number of blue spots is big?

I do find it a little distracting in this scenario. I rarely use full-store search though (and I suspect that will be the case for most people), so IMO it makes the most sense to optimise for the majority case of filtering the domain-matched list. My preference for the latter case is to keep the current highlighting approach, but I'm happy to leave that decision up to you guys.

adtac:
(Busy with an exam for the next couple of days, just thought I should let everybody know. I'll make amendments and push changes once I'm done.)

adtac:
>Or dark grey #b2b2b2

I went with a slightly lighter version to add more contrast with the "pass" text (which is black). IMO the light blue looks a bit out of place in the grey background. Lemme know what you guys think.

>I honestly didn't plan to bring up the topic of highlighting using background color again, so instead of forcing you into changing to use text color alone, let me just ask you this: could you please try this scenario for yourself and tell me if you also feel that it is distracting when the number of blue spots is big?

I agree, but as @erayd said I don't think full store searches are that common. But it's still a valid case neverthenonetheless, so I've made the colour be a bit lighter and less contrasting with the background so that it doesn't standout. I went with `#deeaf3`, but please feel free to suggest alternatives. I still think a less contrasting background is better than no background :)

Rebased on master.

adtac:
Untitled

what it looks like now ^

erayd:
@adtac I think I prefer something a little stronger than that for the highlight colour - I've had a bit of a play, and `#cfecff` feels about right for me. It's not as bold as the original highlight colour, but it has a bit more contrast than `#deeaf3`.

@maximbaz Thoughts?

adtac:
>`#cfecff`

Good call, it's a decent compromise, I agree. Updated.

maximbaz:
I'll try later as well 👍 

By the way, @adtac would you mind if we keep this PR open until after I finish with #61? It will make it easier for me to finish with that quite big feature if I don't have to think about two themes, and then we rebase this PR and continue with the two themes here.

adtac:
No problem. Just ping me here when that is merged and you're good to go.

adtac:
Hey @maximbaz, I haven't been following the project's development much. Any updates on the password generation feature that you wanted merged before this? I'd really like to merge this; looking at the issue page, I imagine a few others would like this too.

maximbaz:
Hey, I apologize for the delay and the blocked state, I'll try to prioritize that work in the following days, I would also really like to get back to zero pending PRs in this project 🙂

adtac:
@maximbaz ping :)

P.S. If you think the password generation feature will take more time than originally planned, I propose merging this patch first; just a suggestion :)

maximbaz:
Apologies again. I agree, at this point let's just get this merged, #61 is taking way too much time than I planned. @erayd could you try again if you like the current colors, as well as do the code review? I'll do the same now.

maximbaz:
For the background of the first entry highlighted by default I propose `#d8dee3`, a bit darker than the current `#dee2e6`, to add a bit of contrast to the blue highlight. Other than that, I'm ok with how it looks. @erayd?

erayd:
@maximbaz I will go over this PR tomorrow.

DamienCassou:
Hi guys. What is the status of this PR?

erayd:
@DamienCassou I'm midway through reviewing it.

DamienCassou:
Erayd <notifications@github.com> writes:
> @DamienCassou I'm midway through reviewing it.

thank you very much!

-- 
Damien Cassou
http://damiencassou.seasidehosting.st

"Success is the ability to go from one failure to another without
losing enthusiasm." --Winston Churchill

maximbaz:
Thanks @adtac for your work!

erayd:
Yes, thank you @adtac, and sorry we took so long!

adtac:
thank you!

PR #123, opened 2019-04-18
Move OTP trigger, include version & action

erayd:
Per request from @maximbaz to move this out of `parseFields()`.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -971,9 +971,9 @@ async function saveSettings(settings) {
  */
 function triggerOTPExtension(settings, action, otp) {
     // trigger otp extension
-    for (let key in otpID) {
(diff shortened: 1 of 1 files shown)


src/background.js
```diff
@@ -971,6 +959,35 @@ async function saveSettings(settings) {
...
+ * @since 3.0.13
+ *
+ * @param object settings Settings object
+ * @param string action   Browserpass action
+ * @param object otp      OTP field data
+ * @return void
+ */
+function triggerOTPExtension(settings, action, otp) {
+    // trigger otp extension
+    for (let key in otpID) {
```
maximbaz: let's use "for of" since we are iterating an array

PR #125, opened 2019-04-18
Prevent duplicate install notifications

maximbaz:
The way Debian's `browserpass-webext` package side-loads Browserpass extension makes Chromium trigger "installed" event every time a browser is restarted. Luckily the localStorage is persisted between restarts, so we can store some marker there, distinguishing the first installation from a subsequent one.

If a user _removes_ the extension, browser will purge everything, so if this user installs it again, they will see a notification again, as it will be a proper first installation.

We could theoretically store anything as a marker, I figure an installation timestamp could serve us well one day (e.g. if we will ever need to find a subset of users who installed the extension withing a certain period of time).

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -1014,12 +1014,15 @@ function onExtensionInstalled(details) {
     };
 
     if (details.reason === "install") {
-        show(
-            "installed",
-            "browserpass: Install native host app",
-            "Remember to install the complementary native host app to use this extension.\n" +
-                "Instructions here: https://github.com/browserpass/browserpass-native"
-        );
+        if (!("installed" in localStorage)) {
+            localStorage.setItem("installed", Date.now());
+            show(
+                "installed",
+                "browserpass: Install native host app",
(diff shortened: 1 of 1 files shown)


src/background.js
```diff
@@ -1014,12 +1014,15 @@ function onExtensionInstalled(details) {
     };
 
     if (details.reason === "install") {
-        show(
-            "installed",
-            "browserpass: Install native host app",
-            "Remember to install the complementary native host app to use this extension.\n" +
-                "Instructions here: https://github.com/browserpass/browserpass-native"
-        );
+        if (!("installed" in localStorage)) {
```
erayd: Does this enumerate over every item in local storage in order to check for the presence of one key?

maximbaz: Judging by the "in (false)" test case I made here, yes, it's considerably slower than using `getItem`. Adapted.

https://jsperf.com/localstorage-in-getitem/1

Issue #62, opened 2019-04-02
Report websites where Browserpass doesn't work


maximbaz:
Please use this thread to report websites where Browserpass doesn't fully work, e.g. Browserpass is unable to fill username or password, or fills wrong form, or autosubmit doesn't work.

I'm not promising to support every single website, but I will investigate every example and if possible improve the detection logic.

I will not however add special cases for certain websites, I don't want Browserpass become a collection of broken websites on the Internet 🙂

532910:
Autosubmit doesn't work for:
https://esia.gosuslugi.ru/
Discourse sites:
https://community.opencellid.org
https://community.letsencrypt.org
https://forum.f-droid.org

maximbaz:
Discourse sites will unfortunately not work until they improve their markup, sadly they are not very interested in doing so: https://github.com/discourse/discourse/pull/7056

532910:
https://ticket.coreboot.org/login
Filed wrong: login goes into openid url, login field leaved unchanged

maximbaz:
esia.gosuslugi.ru works for me, make sure to test in the latest version 😉

532910:
I use СНИЛС login

532910:
version 3.0.3

maximbaz:
I don't have the right credentials so I can't really test a successful login, but I click on СНИЛС, fill it with random credentials and I see the "Войти" button becoming grey, something is refreshing, and then I get "wrong username or password" red error, so as far as I can tell auto-submit actually works? 😕 What do you see exactly? And what browser do you use?

532910:
firefox 66.0.1-1, debian sid
it work for test record with random СНИЛС and password (I see "Введено неверное имя пользователя или пароль")
but for my credentials it just doesn't press "Войти" button

maximbaz:
I have the same behavior for random СНИЛС and password, seems like I won't be able to reproduce because I don't have the correct credentials 😞 
Injection is a bit annoying to debug, but I'll give it another try a bit later.

By the way, sadly there will be cases when autosubmit only works in Chromium but not in Firefox, because we have now an additional way of submitting forms, but Firefox doesn't support this (yet?): https://github.com/browserpass/browserpass-extension/pull/55

It's still an improvement, because in v2 such forms didn't work in both Firefox and Chromium 🙂

532910:
can reproduce for random credentials:
```
% pass gosuslugi.ru    
[...]

PR #126, opened 2019-04-18
Handle inability to inject script to all frames

maximbaz:
Chromium can hang indefinitely trying to inject a script to certain kinds of iframes, such as the one present on `ox.credativ.com` (https://github.com/browserpass/browserpass-extension/issues/62#issuecomment-484020195). The workaround is to wait at most 1 second, and then abort the injection.

Hopefully `1 sec` is more than enough for normal cases and we won't mess with anybody's Browserpass, if we see user reports about "Error: Unable to inject script in the top frame", we must increase this timeout.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -294,6 +294,26 @@ async function dispatchFocusOrSubmit(settings, request, allFrames, allowForeign)
 }
 
 /**
+ * Inject script
+ *
+ * @param object settings Settings object
+ * @param boolean allFrames Inject in all frames
+ * @return object Cancellable promise
+ */
+async function injectScript(settings, allFrames) {
+    const MAX_WAIT = 1000;
+
+    return new Promise(async (resolve, reject) => {
+        setTimeout(reject, MAX_WAIT);
+        await chrome.tabs.executeScript(settings.tab.id, {
+            allFrames: allFrames,
+            file: "js/inject.dist.js"
+        });
+        resolve(true);
+    });
+}
+
+/**
  * Fill form fields
  *
  * @param object settings Settings object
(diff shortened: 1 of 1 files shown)


erayd:
When does this start counting from? If a page is slow to load, will it cause problems?

maximbaz:
It starts counting since the moment you tell Browserpass to fill a form for you, I think in vast majority of cases the page is already loaded when users attempt to fill a form.

maximbaz:
By the way, this is another review that is easier to review with whitespace diff off.

src/background.js
```diff
@@ -293,6 +293,26 @@ async function dispatchFocusOrSubmit(settings, request, allFrames, allowForeign)
...
+ *
+ * @param object settings Settings object
+ * @param boolean allFrames Inject in all frames
+ * @return object Cancellable promise
+ */
+async function injectScript(settings, allFrames) {
+    const MAX_WAIT = 1000;
+
+    return new Promise(async (resolve, reject) => {
+        setTimeout(reject, MAX_WAIT);
```
erayd: This looks problematic. On success, it will both reject *and* resolve, which is probably an error.

src/background.js
```diff
@@ -303,10 +323,17 @@ async function dispatchFocusOrSubmit(settings, request, allFrames, allowForeign)
...
+        await injectScript(settings, false);
+    } catch {
+        throw new Error("Unable to inject script in the top frame");
+    }
+
+    let injectedAllFrames = false;
+    try {
+        await injectScript(settings, true);
+        injectedAllFrames = true;
+    } catch {}
```
erayd: Can you please put an explicit noop comment here. Error suppression with no action should have an explaining comment, both to clarify why, and to prevent one of us accidentally attempting to do something with the error later.

Issue #128, opened 2019-04-19
Error: InvalidStateError: A mutation operation was attempted on a database that did not allow mutations


waldgruber:
### General information

-   Operating system + version: Ubuntu 18.04
-   Browser + version: Firefox 66.0.2 (64-Bit)
-   Information about the host app:
    -   How did you install it? downloaded a pre-built binary
    -   If installed an official release, put a version (`$ browserpass --version`): 
    -   If built from sources, put a commit id (`$ git describe --always`): 3.0.6
-   Information about the browser extension:
    -   How did you install it? webstore,
    -   Browserpass extension version as reported by your browser: 3.0.12
---
### Exact steps to reproduce the problem

1. Go to website login screen, 
2. Enter PGP-password
3. username + password inserted after selection -> login successful
4.no error message is shown
5. Go to another login screen (different website)
6. username + password inserted after selection -> login successful
7. error message is shown "Error: InvalidStateError: A mutation operation was attempted on a database that did not allow mutations"

### What should happen?

no error message shown

### What happened instead?

error message

maximbaz:
Hey, you are a second person telling me about this error. I'll push a bugfix release 3.0.15 now, hope it will help, please tell me if the issue is still present after you update.

waldgruber:
Thank you for the fast fix, it works now without a problem!
Bye-the-way: thank you for a great tool!

PR #129, opened 2019-04-19
Ignore Indexed DB errors

maximbaz:
Fixes https://github.com/browserpass/browserpass-extension/issues/128

Review with whitespace diff disabled

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -194,13 +194,17 @@ async function saveRecent(settings, login, remove = false) {
     }
 
     // save to usage log
-    const DB_VERSION = 1;
-    const db = await idb.openDB("browserpass", DB_VERSION, {
-        upgrade(db) {
-            db.createObjectStore("log", { keyPath: "time" });
-        }
-    });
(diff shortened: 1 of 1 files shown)


src/background.js
```diff
@@ -194,13 +194,17 @@ async function saveRecent(settings, login, remove = false) {
...
+    try {
+        const DB_VERSION = 1;
+        const db = await idb.openDB("browserpass", DB_VERSION, {
+            upgrade(db) {
+                db.createObjectStore("log", { keyPath: "time" });
+            }
+        });
+        await db.add("log", { time: Date.now(), host: settings.host, login: login.login });
+    } catch {
+        // ignore if we cannot write to Indexed DB (e.g. it's readonly in incognito mode)
```
erayd: Where's the logic that checks the reason for an exception here? I assume there are more potential error sources than just being unable to write.

Somewhat related - noting this is a log, and isn't required for any of the actual functionality, is throwing the data away regardless of error source acceptable?

maximbaz: Discussed offline, agreed to update comment and that's it

PR #131, opened 2019-04-19
Fill best matching password entry with Ctrl+Shift+F

### LINKED ISSUE #75: Allow a single shortcut to replace Ctrl+Shift+L followed by Enter
### General information

* Operating system + version: Linux / Arch Linux (64-bit)
* Browser + version: Chromium 67.0.3396.87 (Official Build)
* Information about the host app:
  * How did you install it?
    * Installed via a package manager, downloaded a pre-built binary, compiled yourself? **AUR**
  * If installed an official release, put a version: **2.0.21**
    * In the recent versions it can be obtained with `$ browserpass -v`.
    * If in doubt what version you have, download and re-install the latest version!
  * If built from sources, put a commit id (`$ git describe --always`): 
* Information about the browser extension:
  * How did you install it?
    * Installed via webstore, downloaded a pre-built extension, compiled yourself? **AUR**
  * Browserpass extension version as reported by your browser: **2.0.21**

---

### What should happen?

Currently I use the following combination to automatically input the password if I'm focused on the password field: (Ctrl+Shift+L) + (Enter). This will open the popup and select the best matching password (which is almost always the correct password, thanks to my directory structure).

I would like to combine this into one shortcut that automatically chooses and enters the best fitting password, if any. If there no matching passwords, the extension should do nothing.

maximbaz:
Depends on #130, after I rebase the diff will get slightly simpler

Fixes #75

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -64,6 +64,7 @@ chrome.commands.onCommand.addListener(async command => {
             try {
                 const settings = await getFullSettings();
                 handleMessage(settings, { action: "listFiles" }, listResults => {
+                    const logins = helpers.prepareLogins(listResults.files, settings);
                     // TODO this simulates the conversion and applied sorting algorithm as in popup...
                     let bestLogin = {
                         store: { id: "default" },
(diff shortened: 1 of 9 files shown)


src/background.js
```diff
@@ -57,39 +57,34 @@ chrome.runtime.onMessage.addListener(function(message, sender, sendResponse) {
...
+                    if (bestLogin) {
+                        handleMessage(settings, { action: "fill", login: bestLogin }, () => {});
+                    }
+                });
+            } catch (e) {
+                console.log(e);
+            }
+            break;
+    }
+});
```
maximbaz: The review is huge, but this function is the only real difference, the rest is refactoring.

erayd:
Have rebased on master after merging #130.

Issue #135, opened 2019-04-24
Google Drive spreadsheet triggers high CPU usage in Firefox (a infinite loop with browserpass?)


4llan:
### General information

-   Operating system + version: Windows 10 1809
-   Browser + version: Firefox 66.0.3 (64-bits)
-   Information about the host app:
    -   How did you install it?
        Downloaded a pre-built binary (`browserpass-windows64-3.0.6.zip`)
-   Information about the browser extension:
    -   How did you install it?
        Webstore
    -   Browserpass extension version as reported by your browser:
        3.1.0

### Exact steps to reproduce the problem

1. Open any Google Drive Spreadsheet (https://docs.google.com/spreadsheets/) via drive.google.com

2. Firefox CPU and memory usage goes unusual high

3. `browserpass-windows64.exe` is spawned at least 50x

### What should happen?
The extension should list the available logins.

### What happened instead?
* Before opening the spreadsheet (or opening with browserpass extension disabled) ![Before opening the spreadsheet (or opening with browserpass extension disabled)](https://user-images.githubusercontent.com/2865381/56688485-42f25b00-66af-11e9-8ded-423e19c4385e.png)
* Opening spreadsheet via Google Drive ![Opening spreadsheet via Google Drive](https://user-images.githubusercontent.com/2865381/56688521-58678500-66af-11e9-9271-df22fa2f2b14.png)
* Firefox main proccess high CPU and memory usage, plus the spawning of browserpass ![Firefox main proccess high CPU and memory usage, plus the spawning of browserpass](https://user-images.githubusercontent.com/2865381/56688550-6d441880-66af-11e9-87d5-77c45dc61eb1.png)
* 5min later, the proccesses still running and `firefox.exe` still with high CPU consume ![5min later, the proccesses still running and firefox.exe still with high CPU consume](https://user-images.githubusercontent.com/2865381/56688618-94024f00-66af-11e9-998a-214446a70921.png)
* At least 50 child proccesses ![At least 50 child proccesses](https://user-images.githubusercontent.com/2865381/56688701-bb591c00-66af-11e9-83a2-0e0cef4d1d18.png)
* The badge counts the number of logins available for the domain (google.com) but don't load them ![The badge counts the number of logins available for the domain (google.com) but don't load them](https://user-images.githubusercontent.com/2865381/56688729-cad86500-66af-11e9-898d-2a88d4190c67.png)

[...]

PR #141, opened 2019-04-28
Clear credentials from clipboard after 60 seconds

### LINKED ISSUE #140: Clear password from clipboard after 60 seconds
In contrast to pass, browserpass will not clear the contents of the clipboard some time after a password has been copied. Even though the robust auto-filling logic should keep use of this feature to a minimum, it is still potentially dangerous to keep passwords in the clipboard for longer than the user remembers (in my case, sometimes less than a minute ;-)). It would thus be great to have browserpass clear the password automatically after some time.

I believe that it could even be possible to implement this feature without providing any options, simply by making it get out of the way of the user as much as possible. Taking some ideas from pass, it could for example be done this way:

### Implementation

- When the user copies a password, create a new one-shot chrome.alarm named `alarm-<password>` that will clear the contents of the clipboard after one minute (shortest time possible) if the contents are equal to `<password>`. Alternatively, if exposing the password to even more browser APIs is not wanted, replace `<password>` by `<drastically truncated hash>`.
- If the user copies a new password (maybe the same one, who knows?) within the minute, clear the currently set alarm (if any) before creating a new one.

It is also possible to reset the clipboard to its previous contents instead of clearing it. I personally have never liked that feature, as it is tough to figure out what the contents are after some time without trying to paste them, which somewhat defeats the purpose of the feature. Furthermore, if the user copies multiple passwords right after each other, this approach would leave one of the earlier ones in the clipboard in the end.

I would be glad for feedback on both the general feature and the concrete implementation. Maybe we can figure out something that "just works" and can be added without getting on anyone's nerves.

FabianHenneke:
My implementation for #140. This works well on Chrome, but Firefox somehow hangs on the following line indefinitely:
```js
const alarms = await chrome.alarms.getAll();
```
I don't have any experience with Firefox extensions, do you know why this may happen?

----

Fixes #140

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,17 +139,17 @@ If you want to intentionally disable phishing attack protection and search the e
 
 Note: If the cursor is located in the search input field, every shortcut that works on the selected entry will be applied on the first entry in the popup list.
 
-| Shortcut                                             | Action                                          |
-| ---------------------------------------------------- | ------------------------------------------------ |
-| <kbd>Ctrl+Shift+L</kbd>                              | Open Browserpass popup                           |
-| <kbd>Ctrl+Shift+F</kbd>                              | Fill the form with the best matching credentials |
-| <kbd>Enter</kbd>                                     | Submit form with currently selected credentials  |
-| Arrow keys and <kbd>Tab</kbd> / <kbd>Shift+Tab</kbd> | Navigate popup list                              |
-| <kbd>Ctrl+C</kbd>                                    | Copy password to clipboard                       |
-| <kbd>Ctrl+Shift+C</kbd>                              | Copy username to clipboard                       |
-| <kbd>Ctrl+G</kbd>                                    | Open URL in the current tab                      |
-| <kbd>Ctrl+Shift+G</kbd>                              | Open URL in the new tab                          |
-| <kbd>Backspace</kbd> (with no search text entered)   | Search passwords in the entire password store    |
+| Shortcut                                             | Action                                                |
+| ---------------------------------------------------- | ----------------------------------------------------- |
+| <kbd>Ctrl+Shift+L</kbd>                              | Open Browserpass popup                                |
+| <kbd>Ctrl+Shift+F</kbd>                              | Fill the form with the best matching credentials      |
+| <kbd>Enter</kbd>                                     | Submit form with currently selected credentials       |
+| Arrow keys and <kbd>Tab</kbd> / <kbd>Shift+Tab</kbd> | Navigate popup list                                   |
+| <kbd>Ctrl+C</kbd>                                    | Copy password to clipboard (will clear in 60 seconds) |
+| <kbd>Ctrl+Shift+C</kbd>                              | Copy username to clipboard (will clear in 60 seconds) |
+| <kbd>Ctrl+G</kbd>                                    | Open URL in the current tab                           |
+| <kbd>Ctrl+Shift+G</kbd>                              | Open URL in the new tab                               |
+| <kbd>Backspace</kbd> (with no search text entered)   | Search passwords in the entire password store         |
 
 ### Password matching and sorting
 
@@ -271,7 +271,9 @@ Browserpass extension requests the following permissions:
 | -------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `debugger`           | Only used for "auto-submit" functionality: if all attepts to locate a "Submit" button failed, Browserpass will put focus inside the login form and issue an <kbd>Enter</kbd> keypress. This is only available in Chromium-based browsers, and sadly this permission [cannot be specified as optional](https://developer.chrome.com/apps/permissions) |
 | `activeTab`          | To get URL of the current tab, used for example to determine which passwords to show you by default in the popup                                                                                                                                                                                                                                     |
+| `alarms`             | To set a timer for clearing the clipboard 60 seconds after credentials are copied                                                                                                                                                                                                                                                                    |
 | `tabs`               | To get URL of a given tab, used for example to set count of the matching passwords for a given tab                                                                                                                                                                                                                                                   |
+| `clipboardRead`      | To ensure only copied credentials and not other content is cleared from the clipboard after 60 seconds                                                                                                                                                                                                                                         |
 | `clipboardWrite`     | For "Copy password" and "Copy username" functionality                                                                                                                                                                                                                                                                                                |
 | `nativeMessaging`    | To allow communication with the native app                                                                                                                                                                                                                                                                                                           |
 | `notifications`      | To show browser notifications on install or update                                                                                                                                                                                                                                                                                                   |
--- a/src/background.js
+++ b/src/background.js
@@ -28,6 +28,9 @@ var defaultSettings = {
 
 var authListeners = {};
 
+// the last text copied to the clipboard is stored here in order to be cleared after 60 seconds
+let lastCopiedText = null;
+
 chrome.browserAction.setBadgeBackgroundColor({
     color: "#666"
 });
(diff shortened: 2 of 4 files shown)


src/background.js
```diff
@@ -144,6 +173,36 @@ function copyToClipboard(text) {
         { once: true }
     );
     document.execCommand("copy");
+
+    if (!doNotClear) {
```
maximbaz: there's too much negation for my taste, let's rename the parameter to `clear = true`, and this becomes `if (clear)`

src/manifest-firefox.json
```diff
@@ -26,7 +26,9 @@
     },
     "permissions": [
         "activeTab",
+        "alarms",
         "tabs",
+        "clipboardRead",
```
maximbaz: Please add these to README in the section describing permissions

src/manifest-firefox.json
```diff
@@ -26,7 +26,9 @@
...
         "clipboardWrite",
```
maximbaz: Let's also add to README a note saying that copied credentials will be cleared in 60 seconds.

Nothing fancy, perhaps just update "Available keyboard shortcuts" section:

> <kbd>Ctrl+C</kbd> Copy password to clipboard (will clear in 60 seconds)
> <kbd>Ctrl+Shift+C</kbd> Copy username to clipboard (will clear in 60 seconds)

src/background.js
```diff
@@ -80,6 +80,18 @@ chrome.commands.onCommand.addListener(async command => {
     }
 });
 
+// handle fired alarms
+chrome.alarms.onAlarm.addListener(async alarm => {
+    if (alarm.name.startsWith("clearClipboard-")) {
+        const fingerprint = await calculateFingerprint(readFromClipboard());
```
maximbaz: Because it seems we are guaranteed to always have either 0 or 1 active alarm to clear clipboard, how about we hardcode the alarm name as "clear-clipboard", and then in some global variable keep either `null` or last copied data? This way we don't really need to hash anything as the data stays local to background script, and here you would have something like 

```
if (alarm.name === "clear-clipboard" && readFromClipboard() === lastCopiedData) {
    await copyToClipboard(" ", true);
}
lastCopiedData = null
```

I hope alarms are scoped to our extension, i.e. no other extension is able to read or trigger our alarms, do you know?

src/background.js
```diff
@@ -144,6 +173,36 @@ function copyToClipboard(text) {
...
+        // clear all alarms previously set by this method
+        const alarms = await chrome.alarms.getAll();
+        for (const alarm of alarms) {
+            if (alarm.name.startsWith("clearClipboard-")) {
+                await chrome.alarms.clear(alarm.name);
+            }
+        }
+        // clear clipboard after 1 minute if its content still has the same fingerprint as text
+        const fingerprint = await calculateFingerprint(text);
+        await chrome.alarms.create(`clearClipboard-${fingerprint}`, { delayInMinutes: 1 });
```
maximbaz: I'm not sure this function returns a Promise, and therefore you probably don't need to await it, and this entire function doesn't need to be `async`. Could you confirm?

By the way, you don't need to clear the alarm beforehand:

> Alarm names are unique within the scope of a single extension. If an alarm with an identical name exists, the existing alarm will be cleared and the alarm being created will replace it.

So all you need here is:

```
if (clear) {
    chrome.alarms.create("clear-clipboard", { delayInMinutes: 1 });
    lastCopiedData = text;
}
```

src/manifest-firefox.json
erayd: @maximbaz We should maybe do this in 3.2, because adding this permission will disable it for everybody when the extension auto-updates.

maximbaz: Good point - let's target this for 3.2

@FabianHenneke please update `@since 3.1.2` to `@since 3.2.0` :)

src/background.js
```diff
@@ -80,6 +80,18 @@ chrome.commands.onCommand.addListener(async command => {
...
 });
 
+// handle fired alarms
+chrome.alarms.onAlarm.addListener(async alarm => {
+    if (alarm.name.startsWith("clearClipboard-")) {
+        const fingerprint = await calculateFingerprint(readFromClipboard());
+        // clear the clipboard if the fingerprint of the content hasn't changed
+        if (alarm.name === `clearClipboard-${fingerprint}`) {
+            // setting clipboard to "" is not fully reliable, hence set it to " "
+            await copyToClipboard(" ", true);
```
maximbaz: Why do you think it's not reliable? Works well for me in Chromium and Firefox, plus it doesn't pollute clipboard with a space character, but simply clears it.

src/background.js
FabianHenneke: My idea here was to minimize our dependency on the background page being persistent, which will likely go away in manifest version 3. If you think that it's fine to add features using persistence in the meantime, than I would also gladly go for the easier solution.

maximbaz: Let's go for easier solution, given that we already have persistence for modal HTTP auth and I'm gonna add one more for badge counters to properly solve #135, it's not going away any time soon 🙂

src/background.js
FabianHenneke: It was not reliable when I first implemented it for smart-pass, but that is about two years ago. If it works, it's clearly preferable to copying a space character.

FabianHenneke:
Thanks for the detailed review! I have now addressed all the comments.
It took me a while to figure out the textarea incantations that please both Firefox and Chrome when it comes to pasting clipboard contents, but it seems to work reliably in both browsers now. Could you also verify this?

FabianHenneke:
For the record: https://bugzilla.mozilla.org/show_bug.cgi?id=1364708#c2 is the only code snippet that I found that works.

README.md
```diff
@@ -272,6 +272,7 @@ Browserpass extension requests the following permissions:
...
 | `activeTab`          | To get URL of the current tab, used for example to determine which passwords to show you by default in the popup                                                                                                                                                                                                                                     |
 | `tabs`               | To get URL of a given tab, used for example to set count of the matching passwords for a given tab                                                                                                                                                                                                                                                   |
+| `clipboardRead`      | To ensure only copied credentials and not other content is cleared from the clipboard after 60 seconds                                                                                                                                                                                                                                         |
```
maximbaz: Please also add a note about the `alarms` permission

src/background.js
```diff
@@ -144,6 +158,31 @@ function copyToClipboard(text) {
         { once: true }
     );
     document.execCommand("copy");
+
+    if (clear) {
+        lastCopiedText = text;
+        chrome.alarms.create(`clearClipboard`, { delayInMinutes: 1 });
```
maximbaz: you can change back to regular `"` string here

erayd:
Related to the permissions discussion above - can the `clipboardRead` permission be requested on demand the first time somebody tries to use this feature? That would help avoid a mass-disabling of the extension for users, and would help to clarify its purpose for those who don't read the documentation (which let's face it, will be a large fraction of the user base).

maximbaz:
I'm not sure that's a good idea to add optional permissions at this point, it adds a bit of complexity, potential browser issues, and inconsistency with other permissions that are not optional...

How about we keep them required and instead add an upgrade notification "New permissions added to clear copied credentials after 60 seconds"? We already have a framework for that in `background.js#onExtensionInstalled()`.

FabianHenneke:
Would this work though? I think that the extension would be disabled before the notification can be shown. But of course this would at least explain why after the fact better than the permission dialog by itself.

maximbaz:
It will be disabled for those who install extension via webstore, but as soon as people re-enable it they will instantly understand why we suddenly requested more permissions — sounds acceptable to me, a good compromise between simplicity and usability, let's ask what @erayd thinks 🙂

FabianHenneke:
I implemented @maximbaz's notification proposal in the newest changeset. @erayd, if you would prefer the permission to be optional and know how to make this work in Firefox, I could always revert to the previous version.

erayd:
@FabianHenneke If you aren't able to get optional permissions to behave in Firefox, then let's leave things as-is, with the permission mandatory. Firefox is annoying enough already with its sporadic poor / incomplete API implementations, and this is probably just another such area.

FabianHenneke:
I haven't been able to get optional permissions to work in Firefox, so I would prefer to go with the notification approach.

PR #161, opened 2019-08-31
Add option to ignore entries matching a pattern

erayd:
## What
Adds an `ignore` option to `browserpass.json`. This takes either a string or an array of strings, with each string being a pattern using the standard `.gitignore` syntax. All matching files will be completely ignored by Browserpass.

## Why
Because I have a fair amount of stuff stored in my pass repo which shows up by default in Browserpass, but really shouldn't (e.g. SSH passwords, HTTPS certificates, decryption passwords for associated keys, etc).

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -121,7 +121,7 @@ async function updateMatchingPasswordsCount(tabId) {
             throw new Error(`Unable to determine domain of the tab with id ${tabId}`);
         }
 
-        const files = helpers.ignoreFiles(settings, response.data.files);
+        const files = helpers.ignoreFiles(response.data.files, settings);
         const logins = helpers.prepareLogins(files, settings);
         const matchedPasswordsCount = logins.reduce(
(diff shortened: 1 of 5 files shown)


src/background.js
```diff
@@ -121,7 +121,8 @@ async function updateMatchingPasswordsCount(tabId) {
             throw new Error(`Unable to determine domain of the tab with id ${tabId}`);
         }
 
-        const logins = helpers.prepareLogins(response.data.files, settings);
+        const files = helpers.ignoreFiles(settings, response.data.files);
```
maximbaz: I'd swap the arguments just for consistency with `prepareLogins` where `settings` is the last argument

src/helpers.js
```diff
@@ -261,6 +263,33 @@ function highlightMatches(entry, fuzzyResults, substringFilters) {
     });
 }
 
+/**
+ * Filter out ignored files according to .browserpass.json rules
+ *
+ * @since 3.1.2
```
maximbaz: Probably `3.2.0` together with light theme :)

PR #171, opened 2019-09-26
Send more keyboard events to improve autofill

maximbaz:
Fixes this specific example: https://github.com/browserpass/browserpass-extension/issues/62#issuecomment-529203408

The main trick was to send all those keyboard events not only after `.value` was changed, but also before (!). I also sorted them in the order that I think they are emitted when a real user fills the field.

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -380,14 +380,13 @@
         }
 
         // Focus the input element first
-        var el = find(field, form);
+        let el = find(field, form);
         if (!el) {
             return false;
         }
-        var eventNames = ["click", "focus"];
-        eventNames.forEach(function(eventName) {
+        for (let eventName of ["click", "focus"]) {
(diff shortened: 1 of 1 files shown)


src/inject.js
```diff
@@ -380,28 +380,41 @@
...
             return false;
         }
 
-        // Now set the value and unfocus
+        // Focus the potentially new element again
+        for (let eventName of ["click", "focus"]) {
+            el.dispatchEvent(new Event(eventName, { bubbles: true }));
+        }
+
+        // Send some keyboard events indicating that value modification has started
```
erayd: Can you add a comment here indicating that the key events don't actually have an associated keycode? The behavior makes sense, but if we find a site at some stage which requires it, such a comment will assist with debugging it.

PR #172, opened 2019-09-26
Improve domain matching to reduce false positives

### LINKED ISSUE #170: Single words matching domain extensions
### General information

-   Operating system + version: **Gentoo 17.1**
-   Browser + version: **Chromium 77.0.3865.75**
    -   How did you install it? **Installed via a package manager**
    -   If installed an official release, put a version **3.0.6**
    -   How did you install it? **Installed via Chrome webstore**
    -   Browserpass extension version as reported by your browser: **3.1.1**

---

1. Add directory to pass store containing a single word that might be a domain extension (lets say "**com**")

2. open a URL in browser with this domain extension: **domain.com**

3. get all the passwords existing inside this directory, including all sub directories

### What should happen?

I am not sure if this is by design but it's actually weird to have all the passwords for a word coming up if a domain extension equals that word used as directory name. Giving the number of "funky" domain extensions we have now, it can cause quite some confusion. The other thing is that if you have custom domain extensions like "local" or anything and directories like this.

Wouldn't it be better to have domain names starting to work if there is at least a **.** inside the string? So any **domain.com** works but not **com** itself.

### What happened instead?

Every password inside that directory pops up. This is confusing if you have a directory that has the same name as some existing domain extension.

maximbaz:
@erayd is this what you had in mind for #170?

I also added some comments because this logic becomes quite complex 🙂 
(so complex I made a mistake when writing it, as seen in the git history)

Fixes #170

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -34,8 +34,8 @@ function pathToDomain(path, currentHost) {
 
         // Part is considered to be a domain component in one of the following cases:
         // - it is a valid domain with well-known TLD (github.com, login.github.com)
-        // - it is a valid domain with unknown TLD but the current host is it's subdomain (pi.hole, login.pi.hole)
-        // - it is not a valid domain but the current host matches it EXACTLY (localhost)
+        // - it is a valid domain with unknown TLD but the current host is it's subdomain (login.pi.hole)
+        // - it is or isnt a valid domain but the current host matches it EXACTLY (localhost, pi.hole)
         if (
             t.isValid &&
             ((t.domain !== null && (t.tldExists || currentHost.endsWith(`.${t.hostname}`))) ||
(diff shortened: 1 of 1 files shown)


erayd:
@maximbaz Almost. I don't think we should match for unknown TLD though. Is there a particular reason why we are doing that currently?

maximbaz:
We support hostnames if they match exactly what's currently in the address bar (case 3 in my comment in the code, initially requested for `pi.hole` in https://github.com/browserpass/browserpass-extension/issues/62#issuecomment-482839131 and implemented in https://github.com/browserpass/browserpass-extension/pull/107), it seems natural that we also want to support subdomains for such entries (case 2 in my comment in the code, example of that would be if `login.pi.hole` is currently in the address bar and a user has a `pi.hole.gpg` pass entry).

PR #173, opened 2019-09-26
Reduce number of calls to host app when computing badge counters

### LINKED ISSUE #135: Google Drive spreadsheet triggers high CPU usage in Firefox (a infinite loop with browserpass?)

### LINKED ISSUE #138: "Loading available logins..." get's stuck.
### General information

-   Operating system + version: Windows Server 2019 Standard Edition, Version 1809
-   Browser + version: Firefox Quantum 66.0.3 (64-Bit)
    -   Host app installed via MSI (3.0.6)
    -   Installed via addons.mozilla.org
    -   Browserpass extension version as reported by your browser: 3.1.1
- Using gpg/smartcard for pw decryption

### The Issue
Reproduce:
1. Browse for a while (exact cause unknown)

When using the key combination ctrl+shift+L the popup appears and only shows "Loading available logins..." forever. It seems stuck and will happen from then on on every page. Restarting firefox solves the issue for a while.  
I tried a few things to reproduce it, but it seems pretty random so far. I'll update if new info comes up.

maximbaz:
@erayd what do you think of such naive caching approach? I didn't want to deal with tab management (monitoring when user navigates between tabs, when URL in the tab is changed, when tab is opened or closed, etc.), so instead on _every_ tab event a badge is refreshed. However now the badge counter is cached per hostname, so in majority of cases the code will not attempt to talk to host app.

I recommend reviewing with `?w=1`

Fixes #135 
Fixes #138

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -28,6 +28,8 @@ var defaultSettings = {
 
 var authListeners = {};
 
+var badgeCounters = {};
+
 // the last text copied to the clipboard is stored here in order to be cleared after 60 seconds
 let lastCopiedText = null;
 
@@ -45,10 +47,8 @@ chrome.tabs.onUpdated.addListener((tabId, info) => {
         }
     }
 
-    // refresh badge counter when url in a tab changes
-    if (info.url) {
-        updateMatchingPasswordsCount(tabId);
-    }
+    // redraw badge counter
+    updateMatchingPasswordsCount(tabId);
 });
 
 // handle incoming messages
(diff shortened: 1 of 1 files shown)


src/background.js
```diff
@@ -102,39 +106,53 @@ chrome.runtime.onInstalled.addListener(onExtensionInstalled);
...
+async function updateMatchingPasswordsCount(tabId, forceRefresh = false) {
     try {
-        const settings = await getFullSettings();
-        var response = await hostAction(settings, "list");
-        if (response.status != "ok") {
-            throw new Error(JSON.stringify(response));
+        if (forceRefresh || Date.now() > badgeCache.expires) {
+            badgeCache = {
+                files: null,
+                settings: null,
```
erayd: Perhaps put `await getFullSettings()` here? Seems odd to define this null, and then immediately set it.

src/background.js
```diff
@@ -102,39 +106,53 @@ chrome.runtime.onInstalled.addListener(onExtensionInstalled);
...
  */
-async function updateMatchingPasswordsCount(tabId) {
+async function updateMatchingPasswordsCount(tabId, forceRefresh = false) {
     try {
-        const settings = await getFullSettings();
-        var response = await hostAction(settings, "list");
-        if (response.status != "ok") {
-            throw new Error(JSON.stringify(response));
+        if (forceRefresh || Date.now() > badgeCache.expires) {
+            badgeCache = {
```
erayd: We shouldn't overwrite the cache here. Build a new cache object, and then overwrite once it's complete. That way we avoid any race conditions where something might go to look at the cache, and find it in the middle of a refresh.

src/background.js
```diff
@@ -102,39 +106,53 @@ chrome.runtime.onInstalled.addListener(onExtensionInstalled);
...
+
+            var response = await hostAction(badgeCache.settings, "list");
+            if (response.status != "ok") {
+                throw new Error(JSON.stringify(response));
+            }
+
+            badgeCache.files = response.data.files;
+        }
+
+        if (badgeCache.files === null) {
```
erayd: As above. If we eliminate the race condition, then there's no need for this check.

src/background.js
```diff
@@ -102,39 +106,53 @@ chrome.runtime.onInstalled.addListener(onExtensionInstalled);
...
             const tab = await chrome.tabs.get(tabId);
-            settings.host = new URL(tab.url).hostname;
+            badgeCache.settings.host = new URL(tab.url).hostname;
         } catch (e) {
             throw new Error(`Unable to determine domain of the tab with id ${tabId}`);
         }
 
-        const files = helpers.ignoreFiles(response.data.files, settings);
-        const logins = helpers.prepareLogins(files, settings);
+        // Compule badge counter
```
erayd: s/Compule/Compile/

PR #174, opened 2019-09-27
Avoid live-updating badgeCache object

maximbaz:
Addressing comments in https://github.com/browserpass/browserpass-extension/pull/173

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -31,7 +31,8 @@ var authListeners = {};
 var badgeCache = {
     files: null,
     settings: null,
-    expires: Date.now()
+    expires: Date.now(),
+    isRefreshing: false
 };
 
 // the last text copied to the clipboard is stored here in order to be cleared after 60 seconds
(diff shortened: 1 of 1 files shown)


maximbaz:
Merging as I feel confident this is what you had in mind, if you have comments, do leave them 😉

src/background.js
```diff
@@ -113,24 +114,24 @@ chrome.runtime.onInstalled.addListener(onExtensionInstalled);
...
+            let response = await hostAction(settings, "list");
             if (response.status != "ok") {
                 throw new Error(JSON.stringify(response));
             }
 
-            badgeCache.files = response.data.files;
+            badgeCache = {
+                files: response.data.files,
+                settings: settings,
+                expires: Date.now() + 60 * 1000,
```
erayd: Please don't use magic numbers.

src/background.js
```diff
@@ -113,24 +114,24 @@ chrome.runtime.onInstalled.addListener(onExtensionInstalled);
...
+                isRefreshing: false
+            };
         }
 
-        if (badgeCache.files === null) {
-            return; // badge cache refresh in progress
+        if (badgeCache.isRefreshing) {
```
erayd: Doesn't this condition belong at the very beginning of the function?

maximbaz: You are right, it does, remnants of the previous logic 👍

src/background.js
maximbaz: 👍  will extract

Issue #164, opened 2019-09-05
Treat different ports similarly to subdomains


duckbrain:
### General information

-   Operating system + version: Arch Linux (up to date)
-   Browser + version: Google Chrome 76.0.3809.132 (Official Build) (64-bit) stable
-   Information about the host app:
    -   How did you install it?
     pacman browserpass and browserpass-chrome (aur)
    -   Version 3.0.6
-   Information about the browser extension:
    -   How did you install it?  Chrome Extension Store
    -   Browserpass extension version as reported by your browser: 3.1.1

---

Currently, it appears that the ports are excluded when comparing domain names to include in the  initial drop-down lists.

I have a domain with several ports for various services. My pass tree looks something to the effect of below:

```
├── example.com:2031
│   └── root
├── example.com:2083
│   └── user
```

When I navigate to `example.com:2031` it lists `root` and `user`. I think it would be better to keep the ports separate. Browsers treat different ports as different origins.

Currently subdomains listings include parent domain logins. (Logging in to `www.example.com` lists the logins for `example.com`) I think something like that would be appropriate to have a similar behavior for ports, so logging into `example.com:2031` would include the logins for `example.com`, but not visa-versa, and `example.com` (no explicit port 80 or 443) would not show

maximbaz:
I looked into it, this is a bit tricky because the library we use is not able to parse port, it just strips it entirely.

While I did create a feature request https://github.com/oncletom/tld.js/issues/136, that repo looks abandoned, so I think we will have to implement our in-house solution for this.

PR would be welcome, the relevant place to modify would be [`pathToDomain` function in `helper.js`](https://github.com/browserpass/browserpass-extension/blob/dce3358d7f43531a5eb8db936f17a8539725d91a/src/helpers.js#L27)

duckbrain:
Hmm, I'm not sure port makes sense in tld.js. It seems to be concerned strictly with domains as opposed to other parts of the URI. Because the `parts` are already split on the `/`, parsing out the port separately should be pretty simple: `var portSuffix = parts[key].match(/:[0-9]+$/)[0];` The other issue then become that `URI#hostname` is used (being the domain or the IP address) as opposed to `.host` which is the hostname with the port if applicable.

maximbaz:
Well spotted about `hostname`, you are right we should be using `host` everywhere (e.g. when setting `settings.host`).

Would be nice to have parsing hidden behind a single call to `tldjs.parse()`, but I dont mind either way.

[...]

PR #178, opened 2019-09-28
Port matching

duckbrain:
Adds matching ports with hosts. #164

I moved the match detection (as it is no longer trivial) into `pathToDomain`, which is renamed to `pathToHostInfo`, returns an object `{host: string, isMatch: boolean}`. Still returns null if it's not a valid domain.

Also change `new URL(...).hostname` references to `new URL(...).host`.

Also adds unit testing with Jest, but I can remove that if it's undesired. I also tested in Chromium, where it seems to be working as intended with my password store.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /dist-webstore
 
 /src/node_modules
+/src/coverage
 /src/css
 /src/js
 /src/*.log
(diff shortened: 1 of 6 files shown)


erayd:
Thanks for this. I'd like to hold off on it for now though; I think it would be better to resolve the situation with tldjs etc. first.

Unit testing is generally a good thing, but belongs in a separate PR.

maximbaz:
Big thanks for the PR @duckbrain! In this particular case I'd want us to go with the approach in #179 to get rid of tldjs entirely, but rest assured we will get the port matching implemented for the next release! 😉

PR #179, opened 2019-09-29
Replace tldjs with @browserpass/url, check port

erayd:
## What
 * Replace tldjs with @browserpass/url
 * Only show logins with a port for origins with a matching port

## Why
 * Because tldjs is no longer maintained (#164)
 * Because ports are part of the origin

@maximbaz Do you prefer this approach for port matching, or the one in #178? I added it here because it was easy to do both these tasks at the same time, but if another approach is desired I can adjust the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,11 +161,11 @@ Note: If the cursor is located in the search input field, every shortcut that wo
 
 ### Password matching and sorting
 
-When you first open the Browserpass popup, you will see a badge with the current domain in the search input field:
+When you first open the Browserpass popup, you will see a badge with the current domain name in the search input field:
 
 ![image](https://user-images.githubusercontent.com/1177900/54785353-52046a00-4c26-11e9-8497-8dc50701ddc4.png)
 
-This means that phishing attack prevention is enabled, and Browserpass is only showing you entries from your password store that match this host.
+This means that phishing attack prevention is enabled, and Browserpass is only showing you entries from your password store that match this domain.
 
 In order for Browserpass to correctly determine matching entries, it is expected that your password store follows naming conventions (see [Organizing password store](#organizing-password-store)). In particular your file or folder name must contain a domain with a valid TLD, i.e. not `github.gpg`, but `github.com.gpg`. If an attacker directed you to `https://github.co/login` (notice `.co`), Browserpass will **not** present `github.com` entry in the popup. However if you intentionally want to re-use the same credentials on multiple domains (e.g. `amazon.com` and `amazon.co.uk`), see [How to use the same username and password pair on multiple domains](#how-to-use-the-same-username-and-password-pair-on-multiple-domains).
 
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ If not, repeat the installation instructions for the extension.
 
 Browserpass was designed with an assumption that certain conventions are being followed when organizing your password store.
 
-1. In order to benefit of phishing attack protection, a password entry file, or any of its parent folders, must contain a full domain name (including TLD like `.com`) in their name in order to automatically match a website. However, entries which do not contain such a domain in their name may still be manually selected.
+1. In order to benefit of phishing attack protection, a password entry file, or any of its parent folders, must contain a full domain name (including TLD like `.com`) and optionally port in their name in order to automatically match a website. However, entries which do not contain such a domain in their name may still be manually selected.
 
     File names are not allowed to contain `\` or `/` characters, because both of them are considered to be path separators.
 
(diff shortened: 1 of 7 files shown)


src/helpers.js
```diff
@@ -2,12 +2,12 @@
 "use strict";
 
 const FuzzySort = require("fuzzysort");
-const TldJS = require("tldjs");
 const sha1 = require("sha1");
 const ignore = require("ignore");
+const BrowserPassURL = require("@browserpass/url");
```
maximbaz: For consistency lets use lowercase `P` in `Browserpass` :)

src/helpers.js
```diff
@@ -18,30 +18,30 @@ module.exports = {
...
             continue;
         }
-        var t = TldJS.parse(parts[key]);
+        var info = BrowserPassURL.parseHost(parts[key]);
 
         // Part is considered to be a domain component in one of the following cases:
         // - it is a valid domain with well-known TLD (github.com, login.github.com)
-        // - it is a valid domain with unknown TLD but the current host is it's subdomain (login.pi.hole)
-        // - it is or isnt a valid domain but the current host matches it EXACTLY (localhost, pi.hole)
+        // - it is or isn't a valid domain with unknown TLD but the current host is its subdomain (login.pi.hole)
```
maximbaz: This change in logic is not desired, it really must be a valid domain, otherwise `.password-store/com/example.gpg` will show up on `github.com`

src/helpers.js
```diff
@@ -70,9 +71,13 @@ function prepareLogins(files, settings) {
...
                 allowFill: true
             };
-            login.domain = pathToDomain(storeId + "/" + login.login, settings.host);
+            let domainInfo = pathToDomainInfo(storeId + "/" + login.login, settings.host);
+            login.domain = domainInfo ? domainInfo.hostname : null;
             login.inCurrentDomain =
-                settings.host == login.domain || settings.host.endsWith("." + login.domain);
+                host.hostname == login.domain || host.hostname.endsWith("." + login.domain);
+            if (domainInfo && domainInfo.port && domainInfo.port !== host.port) {
+                login.inCurrentDomain = false;
```
maximbaz: I think we better rename `inCurrentDomain` to `inCurrentHost` not to confuse ourselves later on...

src/helpers.js
erayd: Isn't this just a correction to the comment wording? I didn't think I'd changed the logic here...

src/helpers.js
```diff
@@ -18,30 +17,30 @@ module.exports = {
 /**
  * Get the deepest available domain component of a path
  *
- * @since 3.0.0
+ * @since 3.2.3
  *
  * @param string path        Path to parse
  * @param string currentHost Current hostname for the active tab
- * @return string|null Extracted domain
+ * @return string|null Extracted domain info
```
maximbaz: both params and return docstring need to be adjusted, they are now objects instead of strings.

src/helpers.js
```diff
@@ -18,30 +17,30 @@ module.exports = {
...
-        // - it is or isnt a valid domain but the current host matches it EXACTLY (localhost, pi.hole)
+        // - it is or isn't a valid domain with unknown TLD but the current host is its subdomain (login.pi.hole)
+        // - it is or isn't a valid domain but the current host matches it EXACTLY (localhost, pi.hole)
         if (
-            t.isValid &&
-            ((t.domain !== null && (t.tldExists || currentHost.endsWith(`.${t.hostname}`))) ||
-                currentHost === t.hostname)
+            info.validDomain ||
+            currentHost.hostname.endsWith(`.${info.hostname}`) ||
+            currentHost.hostname === info.hostname
```
maximbaz: I'd swap the last two conditions, at least for me that would sound a bit more natural :)

src/helpers.js
```diff
@@ -70,9 +70,35 @@ function prepareLogins(files, settings) {
...
+                    ? `${pathInfo.hostname}:${pathInfo.port}`
+                    : pathInfo.hostname;
+
+                // check whether extracted path info matches the current origin
+                login.inCurrentHost = host.hostname === pathInfo.hostname;
+
+                // check whether the current origin is subordinate to extracted path info, meaning:
+                //  - that the path info is not a single level (e.g. com, net, local)
+                //  - and that the host ends with that path info
+                if (pathInfo.hostname.includes(".") && host.hostname.endsWith(pathInfo.hostname)) {
```
maximbaz: this must be ``endsWith(`.${pathInfo.hostname}`)``, otherwise `hub.com.gpg` will show up on `https://github.com`.

Change comment also to "- and that the current origin is a subdomain of that path info"

maximbaz:
I fixed the above, but I have one final remark that I'd like us to discuss (so I'll not merge just yet).

How shall we handle `443` and `80` ports?

It sounds to me that if an entry `github.com.gpg` should be considered "portless" (i.e. it shows up on any port), `github.com:8080.gpg` should show up only on `8080` port, therefore `github.com:443.gpg` should only show up on `443` port. The problem is, if you try to put `https://github.com:443` in the browser address bar, browser will cut off the `:443` part.

So as of right now, it is impossible to get `github.com:443.gpg` show up in Browserpass. Would you agree that this is a bug?

If we were to fix it, I think we must refactor again, and replace `settings.host` with `settings.origin` (so to preserve the scheme information), and then add some logic for default `80` and `443` ports.

maximbaz:
Oh and one more thing, we must adjust the [Password matching and sorting](https://github.com/browserpass/browserpass-extension#password-matching-and-sorting) section of the README, I'll make some preliminary changes, feel free to adapt.

erayd:
> So as of right now, it is impossible to get github.com:443.gpg show up in Browserpass. Would you agree that this is a bug?

Yes, I agree that this is a bug.

I would like to think on it a bit more, but my initial inclination is to shim it in @browserpass/url such that it always provides a port. This is straightforward to achieve by simply mapping the http and https schemes to ports 80 and 443 respectively. And then apply that library to wherever else Browserpass happens to be parsing URLs.

I agree that this change may need some more refactoring. Can I think on it a bit, and come back to you tomorrow evening?

maximbaz:
absolutely!

README.md
```diff
@@ -161,16 +161,18 @@ Note: If the cursor is located in the search input field, every shortcut that wo
 
 ### Password matching and sorting
 
-When you first open the Browserpass popup, you will see a badge with the current domain name in the search input field:
+When you first open the Browserpass popup, you will see a badge with the current host in the search input field:
```
erayd: I think we should stick with using "domain" here, and in the section below. It's technically wrong, but IMO obvious, and significantly less confusing. This is particularly noteworthy in light of the fact we also have a "native host" / "host app" component to browserpass.

maximbaz: OK, agreed 👍

PR #183, opened 2019-10-19
Make Ctrl+C copy username if username button is focused

erayd:
## What

If the username button is currently focused, <kbd>CTRL+C</kbd> will copy the username, not the password.

image

## Why

Because the obvious behavior, when the username button is selected, is to copy the username.


maximbaz:
Intuitive UX, I like! 🙂

PR #185, opened 2019-10-19
Allow `data:` scheme for images in CSP

### LINKED ISSUE #184: CSP error when opening options screen
Opening the options screen may result in a CSP violation for an embedded down-arrow image, which appears to be associated with the theme-colour dropdown. ~This appears to be introduced by the `less` CSS preprocessor, as the inline image to which it refers is defined there, but not anywhere in the browserpass source.~ There is no obvious location within either the browserpass source, or any of its dependencies, where this is defined. Less defines a different image with the same prefix.

image

Reproduced on Chrome 76.0.3809.100 (Gentoo Linux) with cbab38f.

maximbaz:
Per https://stackoverflow.com/a/18449556 it seems we need to whitelist `data:` scheme.

This bug appeared because light theme PR for the first time introduced select box in the options screen, and browser styles arrow in the select box using an image constructed with `data:`.

I think there's no harm security-wise in allowing `data:` scheme, what do you think?

Fixes #184

### DIFF
--- a/src/manifest-chromium.json
+++ b/src/manifest-chromium.json
@@ -40,7 +40,7 @@
         "http://*/*",
         "https://*/*"
     ],
-    "content_security_policy": "default-src 'none'; font-src 'self'; img-src 'self'; script-src 'self'; style-src 'self'",
+    "content_security_policy": "default-src 'none'; font-src 'self'; img-src 'self' data:; script-src 'self'; style-src 'self'",
     "commands": {
         "_execute_browser_action": {
             "suggested_key": {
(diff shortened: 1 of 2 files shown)


erayd:
Security-wise I think this is OK. It doesn't locate the root cause, but given the source isn't in our codebase or any of our dependencies, it seems like a reasonable compromise.

There are some remote inclusion vectors in XML which may or may not be exploitable via the SVG parser, but given that you'd still need to compromise the extension first before you could inject it (at which point you'd already have JS execution), I think it's acceptable.

PR #187, opened 2019-10-23
Implement home / end keys for popup

### LINKED ISSUE #186: Add support for Home and End keys for popup navigation
### General information

-   Operating system + version: Arch Linux
-   Browser + version: Firefox v69.0.3
    -   How did you install it?
package manager
        
```
$ browserpass --version
Browserpass host app version: 3.0.6
```
    -   How did you install it?
package manager
        
    -   Browserpass extension version as reported by your browser:
```
$ pacman -Qs browserpass-firefox
local/browserpass-firefox 3.4.1-1
```

---

If you are getting an error immediately after opening popup, have you followed the [Configure browsers](https://github.com/browserpass/browserpass-native#configure-browsers) documentation section?

---

1. Open a site where one has multiple accounts.

1. Press `Ctrl+Shift+L`

1. Press `Up` arrow key

### What should happen?
I'm requesting a feature to go back to the end of the list when pressing the up key if the current selected entry is the top one. It should also go to the top of the list when pressing down key if the current selected entry is the bottom one.

### What happened instead?
Nothing. There is no infinite cycle.

erayd:
## What

Use home / end keys in the popup to scroll to top / bottom respectively.

## Why

 * Logical behavior
 * Closes #186

### DIFF
--- a/src/popup/searchinterface.js
+++ b/src/popup/searchinterface.js
@@ -114,6 +114,15 @@ function view(ctl, params) {
                                 );
                             }
                             break;
+                        case "End": {
+                            let logins = document.querySelectorAll(".login");
+                            if (logins.length) {
+                                let target = logins.item(logins.length - 1);
+                                target.focus();
+                                target.scrollIntoView();
+                            }
+                            break;
+                        }
                     }
                 }
             })
--- a/src/popup/searchinterface.js
+++ b/src/popup/searchinterface.js
@@ -115,11 +115,13 @@ function view(ctl, params) {
                             }
                             break;
                         case "End": {
-                            let logins = document.querySelectorAll(".login");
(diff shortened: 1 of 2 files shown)


src/popup/searchinterface.js
```diff
@@ -114,6 +114,15 @@ function view(ctl, params) {
                                 );
                             }
                             break;
+                        case "End": {
+                            let logins = document.querySelectorAll(".login");
```
maximbaz: Is it possible to check if cursor is currently in the last position in the search box before executing the logic below?  If on `github.com` you type `git<Home><End>` your cursor leaves the search box, could be annoying.

erayd: Yes - good idea. I have implemented this,

PR #188, opened 2019-10-30
Improve autosubmit via Chromium's debugger

maximbaz:
Found this somewhere on the Internet, to my surprise this set of extra properties makes `Enter` simulation... better. Also although not required, I'm now sending another "char" event because that's what that example did - just in case :).

This fixes autosubmit in Chromium on https://grabr.io/

----

@erayd I intentionally kept the "Detached while handling command" errors, the way to get rid of them is to `await` all `chrome.debugger.XXX` calls, but then it becomes very visible that Browserpass enables and disables debugger (like you can see this top bar appearing). It works every time without `await`, and it's completely invisible to the user that we temporarily enable debugger, so I think it's more preferable to keep it this way. What do you think? :)

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -327,14 +327,17 @@ async function dispatchFocusOrSubmit(settings, request, allFrames, allowForeign)
         for (let frame of perFrameResults) {
             if (frame.needPressEnter) {
                 chrome.debugger.attach({ tabId: settings.tab.id }, "1.2");
-                for (let type of ["keyDown", "keyUp"]) {
+                for (let type of ["keyDown", "char", "keyUp"]) {
                     chrome.debugger.sendCommand(
                         { tabId: settings.tab.id },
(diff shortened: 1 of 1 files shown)


erayd:
> ...and it's completely invisible to the user that we temporarily enable debugger...

It's not completely invisible if they have developer mode enabled; it complains loudly on the extensions page:
image

And if the user clicks the 'Errors' button, they see this:
image

How long does the top bar appear for if we wait for it?

maximbaz:
It appears and instantly disappears, but it is visible. More people will be bothered by this than by errors in the console I think :)

erayd:
Agreed, if it's there long enough to be noticeable, then it's a problem. Let's leave things as they are (i.e. accept the error and don't wait for the debugger to finish); anybody who has developer mode enabled is hopefully savvy enough not to be concerned about it, and everybody else won't notice.

PR #189, opened 2019-10-30
Truncate maxlength= fields

erayd:
## What

When filling fields, if a field has `maxlength=` defined, truncate the value to that length prior to filling.

## Why

Because if the value is not truncated, then Browserpass fills the entire value, which deviates from the behavior when the user types the password manually or pastes it (in which case the browser will truncate the value as it's entered).

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -411,6 +411,12 @@
             el.dispatchEvent(new Event(eventName, { bubbles: true }));
         }
 
+        // truncate the value if required by the field
+        let maxLength = el.getAttribute("maxlength");
+        if (maxLength) {
+            value = value.substr(0, maxLength);
+        }
+
         // Set the field value
         let initialValue = el.value || el.getAttribute("value");
         el.setAttribute("value", value);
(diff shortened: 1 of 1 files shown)


src/inject.js
```diff
@@ -411,6 +411,12 @@
             el.dispatchEvent(new Event(eventName, { bubbles: true }));
         }
 
+        // truncate the value if required by the field
+        let maxLength = el.getAttribute("maxlength");
```
maximbaz: do we need `el.maxlength || el.getAttribute("maxlength")` ? I'm not sure why we do this for `value`...

erayd: No, we don't need to do this - but if you'd rather use the property, we can do that :-).

The logic for `value` is different because the object property and the attribute aren't always in sync, at least according to the DOM inspector. `el.value || el.getAttribute("value")` will use the property if set and not empty, otherwise it'll get the attribute. I'm not sure whether they are actually out of sync, or whether the DOM inspector just lies sometimes... but IMO better to play it safe.

maximbaz: I see! I'm okay with any approach, I simply didn't know why we do that for `value` 🙂

PR #192, opened 2019-10-31
Resolve "null is not a valid URL" errors caused by a URL parsing bug in Firefox

### LINKED ISSUE #191: Regression: error on New Tab in Firefox
It is supposed to show a full list of passwords, and it definitely worked like this in the past, not sure when it broke:

image

erayd:
## What

 * Work around parsing issue with URLs in Firefox
 * Adjust code to expect nulls in the origin field

Fixes #191.

## Why

Because Firefox is annoyingly broken, again, and somehow thinks providing `"null"` instead of `null` is reasonable.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -139,7 +139,7 @@ async function updateMatchingPasswordsCount(tabId, forceRefresh = false) {
 
         try {
             const tab = await chrome.tabs.get(tabId);
-            badgeCache.settings.origin = new URL(tab.url).origin;
+            badgeCache.settings.origin = tab.url && new BrowserpassURL(tab.url).origin;
         } catch (e) {
             throw new Error(`Unable to determine domain of the tab with id ${tabId}`);
         }
@@ -590,7 +590,10 @@ async function getFullSettings() {
         let originInfo = new BrowserpassURL(settings.tab.url);
         settings.host = originInfo.host; // TODO remove this after OTP extension is migrated
         settings.origin = originInfo.origin;
-    } catch (e) {}
+    } catch (e) {
+        settings.host = null;
+        settings.origin = null;
+    }
 
     return settings;
 }
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -59,7 +59,7 @@ function pathToInfo(path, currentHost) {
 function prepareLogins(files, settings) {
     const logins = [];
     let index = 0;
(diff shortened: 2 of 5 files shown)


src/background.js
```diff
@@ -139,7 +139,7 @@ async function updateMatchingPasswordsCount(tabId, forceRefresh = false) {
 
         try {
             const tab = await chrome.tabs.get(tabId);
-            badgeCache.settings.origin = new URL(tab.url).origin;
+            badgeCache.settings.origin = tab.url && new BrowserpassURL(tab.url).origin;
```
maximbaz: I don't like that we have to repeat this `tab.url && XXX` logic in many places, easy to miss... Maybe we should think of another approach? e.g. always initialize `BrowserpassURL` object, but make the `origin` field inside of it be null, or some special value, so that we don't have to check for `null` everywhere?

src/background.js
```diff
@@ -139,7 +139,7 @@ async function updateMatchingPasswordsCount(tabId, forceRefresh = false) {
```
maximbaz: Chromium has the same behavior of returning `"null"` 😝 

```
> new URL("chrome://newtab").origin
"chrome://newtab"

> new URL("about:newtab").origin
"null"
```

erayd: Interesting. Wonder why they have the same issue?

Chrome doesn't tend to use about: URLs though, so that may be why it's not been a problem there.

src/background.js
erayd: Are you suggesting that BrowserpassURL should accept `null` as a valid argument? Because that's what your suggestion implies.

maximbaz: Is it ever null? When I tested this, `tab.url` was either `"about:newtab"` in Firefox (and `"chrome://newtab"` in chromium).

erayd: Yes, because that's not the only place we obtain a URL from. In some places we parse the `origin` property of a previously-parsed URL, which may be `null`.

erayd: Also, this specific invocation sometimes has `null` coming from `tab.url`. This showed up in Firefox debugging.

maximbaz: let's rethink it, how about we make `settings.origin` be not a nullable string, but non-nullable BrowserpassURL object that we don't have to ever reparse?

PR #198, opened 2020-02-02
Fix about:XXX support in Firefox

### LINKED ISSUE #191: Regression: error on New Tab in Firefox

maximbaz:
An alternative to #192 

Fixes #191

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -15,7 +15,7 @@
         }
     ],
     "dependencies": {
-        "@browserpass/url": "^1.1.3",
+        "@browserpass/url": "^1.1.6",
         "chrome-extension-async": "^3.3.2",
         "fuzzysort": "^1.1.4",
         "idb": "^4.0.3",
--- a/src/yarn.lock
+++ b/src/yarn.lock
@@ -2,7 +2,7 @@
 # yarn lockfile v1
 
 
-"@browserpass/url@^1.1.3":
+"@browserpass/url@^1.1.6":
   version "1.1.6"
   resolved "https://registry.yarnpkg.com/@browserpass/url/-/url-1.1.6.tgz#53831fe9228ecfba2fec5d0166f3fcbd266e2615"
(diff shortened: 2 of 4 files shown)


src/package.json
```diff
@@ -15,7 +15,7 @@
         }
     ],
     "dependencies": {
-        "@browserpass/url": "^1.1.3",
+        "@browserpass/url": "^1.1.5",
```
maximbaz: this needs 1.1.6

src/yarn.lock
```diff
@@ -2,10 +2,10 @@
 # yarn lockfile v1
 
 
-"@browserpass/url@^1.1.3":
-  version "1.1.3"
-  resolved "https://registry.yarnpkg.com/@browserpass/url/-/url-1.1.3.tgz#8b94896198e7032b80c7d82e836d59e42b769669"
-  integrity sha512-XWhYNZo0BGcyFxkum8g1DeUkFw9QyqcaRTOEinyPSpIubh+aL2VQxbdItitVtS5qkkowqaA5Xt1H2pAqu8ic8w==
+"@browserpass/url@^1.1.5":
+  version "1.1.5"
```
erayd: Should be 1.1.6, based on your version change in the browserpass-url PR.

maximbaz:
@erayd I'll try to resolve conflicts, let's finalize this pr and release a new version, with updated deps and these Firefox fixes?

erayd:
Sounds good :-).

maximbaz:
> Looks good. We may need a validity check to prevent invalid.invalid from displaying in the anti-phishing filter; normally on an about: page we want to see the entire store.

Just checked, Chrome parses URLs like `chrome://extensions` successfully, and Firefox parses URLs like `about:addons` successfully as well, neither produce `invalid.invalid` (what will produce `invalid.invalid` is for example parsing `chrome://extension` in Firefox).

In other words, when things are normal, users should never see `invalid.invalid`, even on internal browser pages. Should we just leave this as it is for now? I'm not sure anyway what's the best way of handling invalid urls at the moment, e.g. when a fill request was issued...

PR #199, opened 2020-02-02
Fix SVG icon in Firefox without allowing insecure CSS

### LINKED ISSUE #190: Regression: badly rendered icon in Firefox
Guess we haven't tested in Firefox after merging light theme PR 😄 

Notice the button that removes domain filter:

image

@532910, you use Firefox, do you also reproduce?

It's supposed to be red and look like this:

image

maximbaz:
All that inline style is useless and can be replaced with just a couple of standard SVG attributes.

Properly fixes #190


maximbaz:
I can observe a minor difference in Chrome, but I think both old and new way look equally fine. The icon in Firefox looks colorful, so it's good.

Feel free to ping me later if you find that it doesn't look reliable and we will revisit this.

PR #213, opened 2020-06-03
Add hotkeys to tooltips

erayd:
## What
Add hotkey hints to tooltips in the popup interface.

## Why
Makes hotkey discoverability easier for users.

### DIFF
--- a/src/popup/interface.js
+++ b/src/popup/interface.js
@@ -81,7 +81,7 @@ function view(ctl, params) {
                         onkeydown: keyHandler.bind(result)
                     },
                     [
-                        m("div.name", [
+                        m("div.name", { title: "Fill username / password | <ENTER>" }, [
                             m("div.line1", [
                                 m(
                                     "div.store.badge",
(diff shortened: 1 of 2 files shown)


src/popup/interface.js
```diff
@@ -108,12 +108,12 @@ function view(ctl, params) {
...
                         m("div.action.copy-password", {
                             tabindex: 0,
-                            title: "Copy password",
+                            title: "Copy password | <CTRL+C>",
                             action: "copyPassword"
                         }),
                         m("div.action.copy-user", {
                             tabindex: 0,
-                            title: "Copy username",
+                            title: "Copy username | <CTRL+SHIFT+C>",
```
maximbaz, line 116: Awesome idea, love it! I would prefer a capitalized way of writing shortcuts, e.g. `<Ctrl+Shift+C>` instead of `<CTRL+SHIFT+C>`, this way it doesn't stand out as much... 🙂

PR #220, opened 2020-07-15
Use binary package name for Debian

boutil:
It is more useful for users to refer to the _binary package_ `webext-browserpass` instead of the _source package name_  for the Debian distribution, as the binary packages are the ones which can be installed by `apt`.

This proposed modification reflects this in the README.
This pull request is similar to https://github.com/browserpass/browserpass-native/pull/96 for browserpass-native.


maximbaz:
Thanks 🙂

Issue #100, opened 2019-04-13
Email address being used instead of password filename


braderhart:
### General information
-   Operating system + version: Arch 5.0.7-arch1-1
-   Browser + version: Chromium 73.0.3683.103
-   Information about the host app: 
    -   How did you install it? Package manager
    -   If installed an official release, put a version (`$ browserpass --version`): 3.0.6
-   Information about the browser extension:
    -   How did you install it? Package manager
    -   Browserpass extension version as reported by your browser: 3.0.9

---
### Exact steps to reproduce the problem
1. Use browserpass on a domain with an `email` field in the password file
2. Password filename matches username to be used

### What should happen?
- Password filename should be used as username

### What happened instead?
- Email address is populated in username field

### Recommendation
- If using email address as username, require it use `login` or `username` fields or be the name of the password file

erayd:
@maximbaz What do you think? I don't have a strong opinion either way.

erayd:
@braderhart We've decided to remove 'email' from the list of fields considered to be a username. Stand by.

PR #225, opened 2020-08-10
Add "email" to list of username field keys in parseFields

Preston12321:
Hey guys. I just made this small change for my personal setup and figured it might be helpful for a wider range of users. This PR just adds `email` and `e-mail` as possible fields to represent the username for a given site.

It seems to me like the use of some sort of `email: me@example.com` line would be fairly common in people's password stores. A lot of my accounts have no `username` line, just an email, but browserpass doesn't handle that situation correctly, instead defaulting to the filename (which is often the URL, not the username/email).

One solution is to instead use `login`/`user`/`username` instead of `email` as the key in those files, but I think the distinction between username and email is important and useful (for clarity, if nothing else).

Not sure how this change might affect autofill, like whether some login forms that accept either only email or only username will be treated correctly (but in my experience, most sites allow you to enter either to sign in). Maybe the extension should draw a distinction between username/email and treat them differently?

I'm willing to help with a different solution if mine is a bit too hacky

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -909,7 +909,7 @@ async function parseFields(settings, login) {
     // parse lines
     login.fields = {
         secret: ["secret", "password", "pass"],
-        login: ["login", "username", "user"],
+        login: ["login", "username", "user", "email", "e-mail"],
         openid: ["openid"],
         otp: ["otp", "totp", "hotp"],
         url: ["url", "uri", "website", "site", "link", "launch"]


maximbaz:
Hey! This actually was the case in the past, see #102

Preston12321:
Oh wow, guess I should've checked. However, I feel like the current behavior can be a bit confusing, and the need to use either the filename or "login/user/username" fields over "email" is restrictive.

What if we added an option like "Allow using email as username" on the extension's settings page? If checked, **then** we add `"email", "e-mail"` to the list of accepted field names.

maximbaz:
To be honest this sounds very niche to warrant a new option and a new release, where an alternative is just to use one of the other field names... 

Also given #100, it definitely seems that some websites do accept only username and not email, so this has to be a per domain option (which we can support, but still, feels a bit too much if you need to add another line in your password entry `useEmailAsLogin: true` instead of just renaming the field)

Preston12321:
The problem with just using one of the other field names for email addresses is that it doesn't solve the issue of autofill in sites that only accept username and not email, since the email might still be filled in erroneously (especially when there is currently no way to differentiate between usernames that are actually just usernames and those that are actually email addresses).

It would also require effort on the part of users that may have their password stores structured differently to refactor to fit this schema, as would any site-specific `useEmailAsLogin` solution. Maybe we should instead be looking at ways to detect login form fields that only allow usernames vs. those that allow both vs. only email. Then we could allow the parsing of email as a completely separate field from the username that only sometimes gets autofilled.

Sounds like this PR isn't going to go anywhere, but I do think this is something that still needs work. Might open an issue instead to continue discussion about this

maximbaz:
> The problem with just using one of the other field names for email addresses is that it doesn't solve the issue of autofill in sites that only accept username and not email, since the email might still be filled in erroneously (especially when there is currently no way to differentiate between usernames that are actually just usernames and those that are actually email addresses).

I actually think the semantics make it quite easy:

- email that is also used to login to a website:

```
login: john@example.com
```

- email that is just stored for reference, but another username is used for login:

```
login: john
email: john@example.com
```

Can you provide an example where it is impossible to achieve what you want?

> It would also require effort on the part of users that may have their password stores structured differently to refactor to fit this schema

That is correct, but it is impossible to support every possible combination of password store structures, that's why we picked a few common ones and [documented them as convention](https://github.com/browserpass/browserpass-extension#organizing-password-store) that Browserpass understands.

I completely understand that the power of `pass` is the complete freedom of organizing, and it is annoying to have to refactor your password store, however that just helps us to reduce the scope of the work and focus on making a solution that works well within the selected variation...

I'm afraid it would be extremely hard to get a nicely working logic that _sometimes_ picks additional fields from users' password entries, both in terms of making it actually work for all the forms on the internet, and for us to reproduce and fix issues.

Preston12321:
> Can you provide an example where it is impossible to achieve what you want?

Preston12321:
> Can you provide an example where it is impossible to achieve what you want?

If you organize your password store the way you suggest, it increases the complexity and takes away some of the script-ability that using pass provides. If you have all of your email addresses that *can* be used to log in as:
```
login: me@example.com
```
but then all others as:
```
email: me@example.com
login: myusername
```
with the actual username saved as `login`, you have two different places the email for a site might be and the `login` line meaning either email or username but no easy way of knowing which it is, short of using a complicated regular expression in all your other scripts to detect email addresses vs usernames. Sure, it could work, but it feels more like a hack than a solution.

Personally, my change to the code has worked fine for me, and I'm happy enough just keeping my forked version and pulling in upstream changes when needed, so I won' t lose any sleep over browserpass's behavior staying the way it is. For anyone who may stumble across this same issue however, I think a fair compromise would be to modify @maximbaz's suggestion a bit so that at least usernames and emails are never stored under the same field name...

 - For emails that *can* be used to log in:
```
login: me@example.com
```
 - For emails that are only for reference:
```
email: me@example.com
```
 - For all usernames:
```
username: myusername
```
OR
```
user: myusername
```

This way, if you only ever use `login` to mean an email that can be used to log in, at least any custom scripts you may write only have to check for the existence of `email` and `login` and not have to make guesses on whether `login` means the email or the username.

No hard feelings and thanks for the awesome project! I hope to contribute more in the future :)

maximbaz:
Thanks for sharing, I think your modification is very interesting and semantically makes sense. 

Out of curiosity, won't it solve your own use-case as well? 🙂 

In your scripts, if you need to know which of your email addresses can be used for login, you query for `login` field, if you want to collect a list of all email addresses ever used on any website, you query for `(login|email)`. Or you primarily want to avoid refactoring your password store?

> No hard feelings and thanks for the awesome project! I hope to contribute more in the future :)

❤️

Preston12321:
> Out of curiosity, won't it solve your own use-case as well? 🙂

Yes and no. It would solve the problem of browserpass giving me the filename (usually URL) instead of my email address when there's no username entry, but at the cost of (in my own opinion) consistency in my password store, along with a lot of future effort.

> Or you primarily want to avoid refactoring your password store?

Yeah, in the sense that refactoring would require me to go through every entry in the store and check whether the site allows email as a login option, as well as checking for all future entries that I add.

Admittedly, the ability to query for only emails that can be used for logins in my scripts does sound enticing, but in my own experience, the number of sites that allow the email or have no username at all far outnumber those that don't. I'd rather optimize for the most common case by making a single modification to this extension and allow all my email entries to have the same consistent field name than modifying all the scripts I use along with adding an extra step to maintaining all the entries in my password store.

But that's just my preference. Everyone visits different sites and sets up their own store differently, so mileage will vary. After all, the point of pass is to be so simple, you can create the setup that works for *you*.

computator:
I would like to add another vote for adding back email, possibly with a toggle in the extension settings. It stopped working awhile ago and now I know why it stopped lol.

For me, I prefer to use whichever is semantically correct for the username field: `user:` (or `login:`) if it's a username, and `email:` if it's an email address. So I would want it to prefer `user:` if both fields are set or if `user:` is the only one, and use `email:` otherwise. If there is a toggle in the extension settings then people who prefer to use the filename instead of email can leave the setting off. or people like me who prefer email over filename can leave it on. For either one you can always specify `user/login:` for whichever entries don't match the global "use email" setting.

maximbaz:
Thanks, the vote is heard. I think an option is not a good way forward, given that browserpass requires a certain password store layout and this is a very specific use-case 🤔 I'd rather prefer to just figure out a good convention to use that would suit everyone...

Here's an idea I'm contemplating: right now we try to find `user/login` in password entry, and if it is not found we fallback eventually to a file name. But if a file name itself is clearly not a username, i.e. if Browserpass knows that it used filename as a domain name (so the same file name cannot be both domain name and username), what if in this case, as a last resort we try `email` field from the password entry? 

This way, users who put usernames in the file name can keep using `email` as a non-login metadata (#100), while at the same time for people who do not do this and name their files simply `github.com.gpg`, the email will be used instead. 

The downside is that if you have an entry `github.com/personal.gpg`, then `personal` will be considered as a username 🤔

computator:
I'm not at all opposed to another solution, and your idea sounds like it would be a good improvement over the current situation. It would solve the issues I personally have in most cases. That said, I'm not seeing why a global option would not work. For anyone who uses the filename as the username they can just leave the option turned off and the `email` field would continue to be treated as non-login metadata like it is at the moment. For others who want `email` to be considered for the username or don't use the filename as a username they can turn the option on. From what I can tell this would give the flexibility to work well for the most people without having to change all the entries in their store to fit a specific layout. For example:

Some potential formats with the setting off:
- ```
  username.gpg:
    email: meta-only@email.com
  ```
- ```
  whatever.gpg:
    user/login: username
    email: meta-only@email.com
  ```
Some potential formats with the setting on:
- ```
  username.gpg:
  ```
- ```
  whatever.gpg:
    user/login: username
  ```
- ```
  whatever.gpg:
    email: username@email.com
  ```
- ```
  whatever.gpg:
    user/login: username
    email: meta-only@email.com
  ```

This would allow more flexibility to work with whatever preexisting conventions people already have in their store or prefer to use.

maximbaz:
A simple reason: I believe a smaller set of options forces us to think creatively and implement something that works really well for everyone within the selected constraints (such as requiring a certain file/folder layout). In my mind this approach has worked really well for us thus far, you will see we have very very small number of options. There are some exceptions, notably when a lot of people argue for opposite things - this particular case with `email` is both quite niche and also I think there is room for out-of-the box thinking to improve a default behavior that will suit everyone 🙂

PR #229, opened 2020-09-01
Add context screen & integrated TOTP support

erayd:
## What
 * Adds a new context screen for displaying pass entries.
 * Adds integrated support for TOTP

## Why
 * Because we need an interface for eventually editing passwords in.
 * Adds ability for users to access data from password entries that Browserpass currently ignores.
 * Resolves browserpass/browserpass-otp#17


src/background.js
```diff
@@ -918,11 +935,7 @@ async function parseFields(settings, login) {
                 Array.isArray(login.fields[key]) &&
                 login.fields[key].includes(parts[0].toLowerCase())
             ) {
-                if (key === "otp") {
-                    login.fields[key] = { key: parts[0].toLowerCase(), data: parts[1] };
-                } else {
-                    login.fields[key] = parts[1];
-                }
+                login.fields[key] = parts[1];
```
maximbaz, line 938: nice!

src/background.js
```diff
@@ -962,6 +975,45 @@ async function parseFields(settings, login) {
...
+    if (settings.enableOTP && login.fields.hasOwnProperty("otp")) {
+        if (login.fields.otp.match(/^otpauth:\/\/.+/i)) {
+            // attempt to parse otp data as URI
+            try {
+                let url = new URL(login.fields.otp.toLowerCase());
+                let otpParts = url.pathname.split("/").filter((s) => s.trim());
+                login.fields.otp = {
+                    raw: login.fields.otp,
+                    params: {
+                        type: otpParts[0],
```
maximbaz, line 989: I think `params.type` is not used anywhere, can we just delete it?

If not, can we force `type: "totp"` here instead of the `// fix default otp type` below? as I understand nothing else is supported besides TOTP type?

If not, I would still like to move the fix closer to the potentially problematic place, I would suggest:

```javascript
            // attempt to parse otp data as URI
            try {
                let url = new URL(login.fields.otp.toLowerCase());
                let otpType = url.pathname.split("/").filter((s) => s.trim())[0];
                if (otpType === "otp") {
                    otpType = "totp";
                }
                login.fields.otp = {
                    raw: login.fields.otp,
                    params: {
                        type: otpType,
```

src/options/interface.js
```diff
@@ -55,6 +55,9 @@ function view(ctl, params) {
             "Automatically submit forms after filling (not recommended)"
         )
     );
+    nodes.push(
+        createCheckbox.call(this, "enableOTP", "Enable support for OTP tokens (not recommended).")
```
maximbaz, line 59: remove the trailing period for consistency with the other checkbox

src/popup/detailsInterface.js
```diff
@@ -0,0 +1,140 @@
...
+    var nodes = [];
+    nodes.push(
+        m("div.part.login.details-header", [
+            m("div.name", [
+                m("div.line1", [
+                    m(
+                        "div.store.badge",
+                        {
+                            style: `background-color: ${storeBgColor};
+                                                color: ${storeColor}`,
```
maximbaz, line 73: there is one extra space on this line... 🙈

src/background.js
```diff
@@ -750,6 +742,28 @@ async function handleMessage(settings, message, sendResponse) {
...
+                        status: "error",
+                        message: "Unable to copy OTP token",
+                    });
+                }
+            } else {
+                sendResponse({ status: "error", message: "OTP support is disabled" });
+            }
+            break;
+
+        case "details":
```
maximbaz, line 764: I would rename it to "getDetails" in spirit of later addition of "saveDetails" similar to settings.

maximbaz (commented):
Looks extremely good!

src/background.js
erayd: We shouldn't be getting rid of it, because we need to check it - otpauth URLs can contain other types (e.g. HOTP, which we don't support). Can change where we do that though.

erayd: Have moved it into a ternary.

maximbaz (approved):

PR #233, opened 2020-09-11
Add entry editing

erayd:
## What
 * Add capability & UI to edit password entries

## Why
 * Because it's useful


maximbaz:
I believe we have captured everything from this one, closing.

Issue #251, opened 2021-01-05
OTP does not work since main integration


be-neth:
### General information

-   Operating system + version:

I use browserpass from Archlinux repository on Archlinux.

-   Browser + version:

Firefox archlinux version 84.0.1-1.

-   Information about the host app:
    -   How did you install it?

Via archilnux package manager.

 -   If installed an official release, put a version (`$ browserpass --version`):

« Browserpass host app version: 3.0.7 » 

-   Information about the browser extension:
    -   How did you install it?

Via Archlinux package manager.

   -   Browserpass extension version as reported by your browser:

Firefox report browserpass extension version 3.7.1.
(community/browserpass-firefox 3.7.1-2 (firefox-addons) [installed])

---

In the past, I use the browserpass OTP extension for generating my OTP and it work fine.
After the integration into the main browserpass addon, I never succeed to have an OTP generated by the extension. I use `pass otp` from command line as a workaround to generate code but it becomes a little bit annoying.

I have enabled option in addon preference to allow OTP generation. But I never see any difference at all.
How can I debug that ?

Thanks for your help,

maximbaz:
Hey, could you please post an example password entry that you use (with redacted info of course) and a screenshot of the extension when you press `>` button on it? 

If I remember correctly, after you fill username/password, OTP token will automatically be copied to clipboard (just like it was in the separate extension), but you can also reopen browserpass popup, go to `>` for details and you should see the token there.

be-neth:
Pass store example for github.com :

~~~
mygithubpassword
url: *.github.com/login
login: be-neth
otpauth://totp/GitHub:be-neth?secret=SECRET&issuer=GitHub
~~~

> screenshot of the extension when you press > button on it?

Interesting, I never use this button before, and it seems to not work on my environment. It trigger as usual the GPG pinentry program to give the PIN code for my Nitrokey (I use a gpg key, nitrokey pro 2) but after returning from pinentry program, nothing is shown on firefox and nothing is copied to my clipboard.

I use Wayland environment with sway and wlroots as wayland compositor.

be-neth:
Ok I feel stupid now…

> If I remember correctly, after you fill username/password, OTP token will automatically be copied to clipboard

You are right, I never check that OTP code was automatically generated and copied to my clipboard. It works like that, it is fine for me.
We can close this issue, thanks for your time.

[...]

PR #252, opened 2021-01-07
README: Add OTP usage

be-neth:
After discussion on issue #251, adding info about OTP usage.

I have take some working from the original `browserpass-otp` [Readme.md](https://github.com/browserpass/browserpass-otp/blob/master/README.md) extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,9 @@ In order to use Browserpass you must also install a [companion native messaging 
     -   [Password store locations](#password-store-locations)
 -   [Options](#options)
     -   [A note about autosubmit](#a-note-about-autosubmit)
-    -   [A note about OTP](#a-note-about-otp)
+    -   [OTP](#otp)
+        -   [A note about OTP](#a-note-about-otp)
+        -   [OTP Usage](#otp-usage)
 -   [Usage data](#usage-data)
 -   [Security](#security)
 -   [Privacy](#privacy)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -265,6 +269,12 @@ There are valid scenarios for using `pass-otp` (e.g. it gives protection against
 
 If you still want the OTP support regardless, you may enable it in the Browserpass settings.
 
+#### OTP usage
```
maximbaz, line 272: I would move this to the end of `Usage` section, so we have `Usage > OTP` and `Options > A note about OTP`

README.md
```diff
@@ -265,6 +269,12 @@ There are valid scenarios for using `pass-otp` (e.g. it gives protection against
...
+
+TOTP seeds may be provided either as an otpauth URL (e.g. `otpauth://totp/Example:alice@google.com?secret=JBSWY3DPEHPK3PXP&issuer=Example`) or as a plain seed (e.g. `totp: JBSWY3DPEHPK3PXP`). Please note that the plain form is unsuitable for any TOTP implementation that does not use a period of 30 seconds and a length of 6 digits.
+
+The OTP code may be accessed by clicking on the Browserpass popup « Open Details ». In addition, the generated OTP code will be copied to the clipboard immediately after Browserpass is invoked, unless Browserpass is already using the clipboard for something else.
```
maximbaz, line 276: This might sound too vague if we say "for something else", how about we say "will be copied to the clipboard immediately after the login form is filled"?

maximbaz, line 276: I think the primary usage is the clipboard, not "Open Details" (as it requires yet another decryption of the entry), so I would prefer to swap them, first say about clipboard, and then in addition the code can also be seen in details.

maximbaz (commented):
Many thanks for your help, makes total sense to document this! 🙂

Issue #259, opened 2021-03-13
Is it possible to forget a password in a given website?


JPUnD:
Maybe this is already supported, but I can't figure out how to make it forget a password that was used in a given website by accident. 
Any pointers?

maximbaz:
There is no UI for this yet, although I think it's something we wanted to eventually add to the context menu.

There is "Clear usage data" in extension options if you are okay with deleting _all_ usage data... 

Technically it is possible to delete a single entry (we store it in localstorage), but it's stored in a [hashed format](https://github.com/browserpass/browserpass-extension/blob/77f6d18f8ea92e68fc0c020cb4065283e45a78c2/src/background.js#L226), so it will be a bit tricky to reverse engineer to find the correct entry... But if you want to try, either put a breakpoint there and login again with the incorrect entry (to see it's hash), or calculate it by hand: if you didn't configure a password store in the settings, its name is "default", origin is the first part of the URL, e.g. " https://github.com" on github, and entry is what you see in browserpass popup, eg. "personal/github.com".

Local storage of the extension can be accessed e.g. in Chromium via settings > extensions > Browserpass > background page > Application tab > local storage, and look there for "recents" key.

Hope that helps for now 🙂

JPUnD:
I'll have to work with it, but maybe it's easier to delete the file and just create it again.

I think this could be a good feature request, should i be submitting it somewhere? (that's all i can do with my zero knowledge on java script)

maximbaz:
Yeah just hitting the "Clear usage data" might be the simplest way to go.

This ticket will serve well as a feature request 👍

PR #260, opened 2021-03-15
Add button to clear history for one item

erayd:
## What
 - Add a button to entries with usage history to clear the history for that item.

## Why
 - Implement feature request from #259.


src/background.js
```diff
@@ -214,23 +214,30 @@ function readFromClipboard() {
...
-    settings.recent[sha1(settings.origin + sha1(login.store.id + sha1(login.login)))] =
-        login.recent;
 
     // save to local storage
     localStorage.setItem("recent", JSON.stringify(settings.recent));
 
-    // a new entry was added to the popup matching list, need to refresh the count
-    if (!login.inCurrentHost && login.recent.count === 1) {
+    // an entry was added / removed to the popup matching list, need to refresh the count
+    if (!login.inCurrentHost) {
```
maximbaz, line 240: should we maybe use `&& login.recent.count < 2`? in end end badge count changes only when you have just used a new entry, or when you removed it...

src/popup/interface.js
```diff
@@ -61,69 +61,91 @@ function view(ctl, params) {
...
+                                                  title: "Forget recent usage here",
+                                                  onclick: (e) => {
+                                                      e.stopPropagation();
+                                                      result.recent.when = 0;
+                                                      document
+                                                          .querySelector(
+                                                              ".part.search input[type='text']"
+                                                          )
+                                                          .dispatchEvent(new Event("input"));
+                                                      result.doAction("clearRecent");
```
maximbaz, line 123: Pressing on the "Clear usage" button for some reason requires me to touch Yubikey, not sure why, but it definitely shouldn't, as I am not intending to decrypt the password entry... And if I don't touch Yubikey, then the recent will not be cleared 🤔

src/popup/icon-eraser.svg
```diff
@@ -0,0 +1,10 @@
+<svg id="outline" height="512" viewBox="0 0 512 512" width="512" xmlns="http://www.w3.org/2000/svg">
```
maximbaz, line 1: Add ` stroke="#c4c4c4" fill="#c4c4c4"` to the `<svg` tag, otherwise the icon is black:

image

src/popup/interface.js
```diff
@@ -61,69 +61,91 @@ function view(ctl, params) {
...
             "div.logins",
-            this.results.map(function (result) {
-                const storeBgColor = result.store.bgColor || result.store.settings.bgColor;
-                const storeColor = result.store.color || result.store.settings.color;
+            this.results
+                .filter((result) => {
+                    return !this.currentDomainOnly || result.inCurrentHost || result.recent.when;
+                })
+                .map(function (result) {
+                    console.log(result);
```
maximbaz, line 69: Probably no need for keeping `console.log` 🙂

max-baz (commented):
Nice idea to place the "clear" button directly in the popup!

ieggel:
HI guys,
When will this be merged?

erayd:
@ieggel It probably won't be, as there's another PR currently under review that's making significant changes to the UI rendering. The feature will need to be re-implemented on top of that.

max-baz:
Closing for cleanup purposes, we can start fresh PR if we ever want to revisit this.

PR #261, opened 2021-04-09
Honor store-specific enableOTP setting

shinenelson:
The [Options] allow for the `enableOTP` feature to be specified within
the `.browserpass.json`, but the feature is not enabled when the
option is specified within `.browserpass.json` and not globally.

This was because the check for the setting was only looking for the
global setting - `settings.enableOTP` and did not check for the
`.browserpass.json` setting ( `settings.stores[*].settings.enableOTP` ).

This changeset fixes that issue and properly honors the `enableOTP`
setting specified within `.browserpass.json` with or without the
global `enableOTP` setting.

[Options]: https://github.com/browserpass/browserpass-extension#options

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -743,7 +743,7 @@ async function handleMessage(settings, message, sendResponse) {
             }
             break;
         case "copyOTP":
-            if (settings.enableOTP) {
+            if (settings.enableOTP || settings.stores[message.login.store.id].settings.enableOTP) {
                 try {
                     if (!message.login.fields.otp) {
                         throw new Exception("No OTP seed available");
@@ -982,7 +982,10 @@ async function parseFields(settings, login) {
     }
 
     // preprocess otp
-    if (settings.enableOTP && login.fields.hasOwnProperty("otp")) {
+    if (
+        (settings.enableOTP || settings.stores[login.store.id].settings.enableOTP) &&
+        login.fields.hasOwnProperty("otp")
+    ) {
         if (login.fields.otp.match(/^otpauth:\/\/.+/i)) {
             // attempt to parse otp data as URI
             try {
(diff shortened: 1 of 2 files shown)


max-baz (approved):
Thanks, looks good to me, @erayd what do you think?

src/background.js
```diff
@@ -743,7 +743,7 @@ async function handleMessage(settings, message, sendResponse) {
             }
             break;
         case "copyOTP":
-            if (settings.enableOTP) {
+            if (settings.enableOTP || settings.stores[message.login.store.id].settings.enableOTP) {
```
erayd, line 746: Please use `getSetting()`.

src/background.js
```diff
@@ -982,7 +982,10 @@ async function parseFields(settings, login) {
     }
 
     // preprocess otp
-    if (settings.enableOTP && login.fields.hasOwnProperty("otp")) {
+    if (
+        (settings.enableOTP || settings.stores[login.store.id].settings.enableOTP) &&
```
erayd, line 986: Please use `getSetting()`.

src/popup/detailsInterface.js
```diff
@@ -103,7 +103,8 @@ function view(ctl, params) {
             ]),
             (() => {
                 if (
-                    this.settings.enableOTP &&
+                    (this.settings.enableOTP ||
+                        this.settings.stores[login.store.id].settings.enableOTP) &&
```
erayd, line 107: Please use the login-scoped settings; it doesn't need to be dereferenced from the higher scope again here.

erayd (changes requested):
Nice catch, thanks 🙂.

shinenelson:
Ping @erayd if the review request was missed

erayd:
> Ping @erayd if the review request was missed

Not missed, just short on time. I'm expecting to have some time to spend working on Browserpass during the coming week, which will include this PR.

PR #275, opened 2021-08-23
Remove the warning about outdated debian packages

jayme-github:
Debian has version 3.7.2 imported to stable, so it's pretty up to date now.
This also changes the links to point to debian stable.


maximbaz:
This is awesome news, thanks!

PR #298, opened 2022-06-27
Mention restricted domains workaround for firefox in the readme

Benjamin-L:
Turns out the set of domains that webextensions are disabled for is configurable by the user, which should probably be mentioned in the readme rather than just going "nothing can be done".


maximbaz:
I didn't know about it, very cool, thanks for sharing!

PR #308, opened 2022-11-06
fix `enableOTP` handling to match docs: prioritize store, then extension config

uninsane:
the README.md says:

> Browserpass allows configuring certain settings in different places places using the following priority, highest first:
> 
> 1. Options defined in specific `*.gpg` files, only apply to these password entries:
>     - `autoSubmit`
> 2. Options defined in `.browserpass.json` file located in the root of a password store:
>     - `autoSubmit`
>     - **`enableOTP`**
>     ...
> 3. Options defined in browser extension options:
>     - Automatically submit forms after filling (aka `autoSubmit`)
>     - Enable support for OTP tokens (aka **`enableOTP`**)

this makes it sound like one should be able to set `enableOTP = true` inside a store's config and leave it unset at the extension level, which sounds intuitive. but when i try that on master i'm not shown any token. unless i've missed something, the code only attempts to read this from the extension-level settings. this PR fixes that to match the behavior of the readme.

tested manually:
- [x] check OTP box in browser config, omit the field in `.browserpass.json` -> shows token
- [x] check OTP box in browser config, `"enableOTP": false` in store -> no token
- [x] leave OTP box unchecked in browser config, set `"enableOTP": true` in the store level -> shows token
- [x] leave OTP box unchecked in browser config, omit `enableOTP` from the store config, set `enableOTP: true` in a secret -> no token

i'm a novice JS dev, so if i'm breaking idioms with my approach here don't hold back in setting me straight 😉

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -11,11 +11,31 @@ const BrowserpassURL = require("@browserpass/url");
 module.exports = {
     prepareLogins,
     filterSortLogins,
+    getSetting,
     ignoreFiles,
     makeTOTP,
 };
 
 //----------------------------------- Function definitions ----------------------------------//
+
+/**
+ * Get most relevant setting value
+ *
+ * @param string key      Setting key
+ * @param object login    Login object
+ * @param object settings Settings object
+ * @return object Setting value
+ */
+function getSetting(key, login, settings) {
+    if (typeof login.settings[key] !== "undefined") {
+        return login.settings[key];
+    }
+    if (typeof settings.stores[login.store.id].settings[key] !== "undefined") {
+        return settings.stores[login.store.id].settings[key];
+    }
+
+    return settings[key];
+}
 
 /**
  * Get the deepest available domain component of a path
--- a/src/background.js
+++ b/src/background.js
@@ -450,7 +450,7 @@ async function fillFields(settings, login, fields) {
     // build focus or submit request
     let focusOrSubmitRequest = {
         origin: new BrowserpassURL(settings.tab.url).origin,
-        autoSubmit: getSetting("autoSubmit", login, settings),
+        autoSubmit: helpers.getSetting("autoSubmit", login, settings),
         filledFields: filledFields,
     };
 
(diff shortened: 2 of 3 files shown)


src/helpers.js
```diff
@@ -8,15 +8,44 @@ const hash = require("hash.js");
...
+ * Get most relevant setting value
+ *
+ * @param string key      Setting key
+ * @param object login    Login object
+ * @param object settings Settings object
+ * @param string scope    SettingScope value to specify where to search for the setting (optional)
+ * @return object Setting value
+ */
+function getSetting(key, login, settings, scope = SettingScope.Login) {
+    if (scope === SettingScope.Login && typeof login.settings[key] !== "undefined") {
```
maximbaz, line 39: I can see why you added `SettingScope` config here, but it's actually not needed - the semantics of this function here is that it _always_ tries to find most relevant setting, and then bubbles up in the priority list. 

It is [here](https://github.com/browserpass/browserpass-extension/blob/0fd15fddf82995304b55daab6a40bdd12fd35330/src/background.js#L911-L913) that we define which settings are allowed to be defined inside a login entry. So this way we never need to think what "scope" to call this function on, instead we simply define in that place what settings should be parsed out of a login entry.

maximbaz (changes requested):
Thanks for catching an inconsistency! 👍 You are on a right track, and I think the code change can be even slightly simpler, please see the comment below.

uninsane:
fixed, thanks for linking to the code path! tested just with and without `enableOTP` in the store settings this time, assuming if that works it should all work.

maximbaz (approved):
Thanks! I'll let @erayd have a look as well, but looks good to me

uninsane:
@maximbaz any movement on this? i'd like to move onto other browserpass work before life gets busy, but i'm not the type to maintain a dev branch that diverges from upstream in more than one direction at a time.

maximbaz:
Sorry for the delay and that it blocks the further development, @erayd what do you think about this, would you have time to have a look here?

erayd (approved):
Thanks - much appreciated 👍

erayd:
@maximbaz I missed this one sorry - I've taken a look, and I'm happy for it to be merged from a code standpoint. I haven't actually tested it - but if you are happy with the behavior, then feel free to merge 🙂.

maximbaz:
With a reservation of me not being the typical user of OTP functionality in Browserpass, I think the behavior did make sense to me, and I'm happy that we fixed an inconsistency and gained the more intuitive behavior 🙂 Let's merge, and if you spot a regression later, we can always revisit 👍

PR #328, opened 2023-10-03
Fix popup.less layout issues for Chromium browsers

mcexit:
Fixes layout issues on Chromium browsers due to LayoutFlexNewColumnAlgorithm.

### DIFF
--- a/src/popup/popup.less
+++ b/src/popup/popup.less
@@ -36,6 +36,7 @@ body {
     min-width: 260px;
     overflow-x: hidden;
     white-space: nowrap;
+    width: min-content;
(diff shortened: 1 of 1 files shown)


maximbaz (approved):
I confirm it fixes the issue on Chromium, thanks! There's no functional difference for Firefox, since it didn't have the issue in the first place (perhaps it defaults to `width: min-content`). Thanks!

maximbaz:
@erayd could you ping me on Signal / Wire / Email when you have some time? Let's chat about perhaps cutting a new release some time soon?

PR #341, opened 2024-01-31
Add feature: highlight the form fields Browserpass will fill

erayd:
Leverages the matching logic used for the icon badge to determine if Browserpass has credentials for the tab, and if so injects the standard script & uses it to highlight the fields that will be filled if filling is triggered.

@maximbaz What do you think - is this something we want? Upside is it's easier to see what Browserpass will do, downside is some websites use the `outline` styling for their own purposes already.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -43,7 +43,7 @@ chrome.browserAction.setBadgeBackgroundColor({
 });
 
 // watch for tab updates
-chrome.tabs.onUpdated.addListener((tabId, info) => {
+chrome.tabs.onUpdated.addListener(async (tabId, info) => {
     // unregister any auth listeners for this tab
     if (info.status === "complete") {
         if (authListeners[tabId]) {
@@ -53,7 +53,25 @@ chrome.tabs.onUpdated.addListener((tabId, info) => {
     }
 
     // redraw badge counter
-    updateMatchingPasswordsCount(tabId);
+    let matchCount = await updateMatchingPasswordsCount(tabId);
+
+    // highlight the default login & secret fields
+    if (matchCount) {
+        try {
+            await chrome.tabs.executeScript(tabId, {
+                allFrames: true,
+                file: "js/inject.dist.js",
+            });
+            await chrome.tabs.executeScript(tabId, {
+                allFrames: true,
+                code: `window.browserpass.highlightLoginFields();`,
+            });
+        } catch (e) {
+            // errors here mean that the browser denied the script injection, typically
+            // because the tab is not allowed to run content scripts (e.g. chrome://
+            // pages etc). We can safely ignore these errors.
+        }
+    }
 });
 
 // handle incoming messages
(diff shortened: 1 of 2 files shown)


erayd:
Known issue with it that I'm not yet sure how to solve, is it will highlight other fields on those sites as well, not just the login ones. This is because our selector rules assume that the user has triggered a fill request, and that therefore there is *something* on the current page that needs to be selected. It doesn't handle the "this isn't actually a login page" case well.

Applying that same logic to non-login pages of a site means that Browserpass gets increasingly more desperate to find something relevant, and may eventually end up highlighting something like a search input.

maximbaz:
Interesting! Just a few quick thoughts (I haven't tried yet, so I might say something way off): 

- What about highlighting fields, only if password field was discovered? Should solve the issue with highlighting search bar, at the cost of not highlighting username field when the form only has username field
- Alternatively, what if we trigger highlight only when popup was opened?
- Alternatively, what if we have a separate shortcut for it, that you only use when you are in doubt? (I don't like this too much)

erayd:
> ...at the cost of not highlighting username field when the form only has username field

I did consider this, but I'm not terribly happy with it due to the resulting inconsistency in behaviour between the highlight logic and the fill logic. I'm of the opinion that this feature is useful if (and only if) we can get it to behave and be consistent with the fill logic; if we can't get that consistency then we shouldn't merge it.

> ...what if we trigger highlight only when popup was opened?

This would be a good part-solution. It ensures consistency, but at the cost of not having any hinting available for those who fill via the hotkey (which is a very common use-case).

> ...what if we have a separate shortcut for it, that you only use when you are in doubt? (I don't like this too much)

I don't like this either. With that said, if we were to implement as a "only when the popup is opened" feature, then we can simply rely on that to trigger it (and thereby the CTRL+SHIFT+L hotkey), so I don't think we need an entirely new hotkey for it.

maximbaz:
Another most-certainly-bad-idea is to make `Ctrl+Shift+F` on the first keypress just highlight the fields, and on the second keypress fill-in them - but it will mess with folks' muscle memory for sure...

I want to use it for a bit longer, right now I'm a bit torn - on the one hand, I like that it's not very obtrusive, on the other hand it still is a bit distracting (at least in the beginning) - even as I browse on Github, it shows up in different search bars, and my attention gets focused there for a second because I'm not used to seeing it there...

image

image

maximbaz:
I think realistically we will receive some backlash if we don't do _something_ about those search bar being highlighted, for example in the recent days I had to visit github, google mail and google chat often, and I couldn't really get used to stop paying attention to the highlighted search bar...

erayd:
I agree. We can't merge it unless we have a solution to that problem.

I'm struggling to think of one though.

max-baz:
Closing for cleanup purposes, we can start fresh PR if we ever want to revisit this.

Issue #340, opened 2024-01-25
Firefox 122.0 no longer displays OTP codes


morciej:
### General information

-   Linux Mint **21.3** Virginia
-   Firefox **122.0**
-   Information about the host app:
    -   Installed via a package manager
    -   Browserpass host app version **3.1.0** (downloaded as binary)
-   Information about the browser extension:
    -   How did you install it?
        Installed from Extensions
    -   Browserpass extension version as reported by your browser: **3.8.0**

---

If you are getting an error immediately after opening popup, have you followed the [Configure browsers](https://github.com/browserpass/browserpass-native#configure-browsers) documentation section? Yes I did/**N/A**

---

### Exact steps to reproduce the problem

1. Upgrade to Firefox 122.0

2. Go to any website requiring OTP that you have account on (e.g. Github)

3. Open the extension and click on **>**

### What should happen?

OTP number and seconds bar should appear

### What happened instead?

"View Credentials" without OTP appears

I tried to disable and re-enable the OTP functionality a couple times to be sure. I also re-issued Make command for the host app. No luck.

I'll try to roll back the browser next.

morciej:
Confirmed by downloading and running extension in Firefox 121.0

raxod502:
> working for me some time earlier this year, but later than this issue was created

Agreed, it was working for me up until just today when I upgraded from Firefox `132.0.2+build2-0ubuntu0.24.04.1~mt1` to `133.0.3+build1-0ubuntu0.24.04.1~mt1` on Ubuntu 24.04. Also broken on Ubuntu 22.04 when I performed the same Firefox version upgrade (132 to 133).

The details menu for a password entry now shows the raw OTP string exactly as displayed in the password file contents, instead of the rotating OTP code.

As with previous commenter, command-line usage remains functional and is a workaround.

I did notice that Ubuntu has moved from `pass-extension-otp` to `pass-otp` as the name of the Pass extension, but I don't think the browser extension actually uses the command-line tool to generate OTPs (if it did, https://github.com/browserpass/browserpass-extension/issues/358 wouldn't happen -_-), so I don't think that's the issue.

PR #361, opened 2024-12-17
Fix otp in firefox v122-132

raxod502:
close https://github.com/browserpass/browserpass-extension/issues/340

see explanation in https://github.com/browserpass/browserpass-extension/issues/340#issuecomment-2547639403

tested by running `make firefox` and then load temporary addon in firefox, it resolves the issue for me


maximbaz:
Fantastic work, thanks for your help! I've included specific FF versions that I could infer from the linked issue :)

PR #379, opened 2025-05-08
Password templates

cab404:
image
image

The interface is bad, but it's a lot more usable than not having it. 

Didn't test on Chrome — will be glad if someone tries it.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -20,6 +20,7 @@ var defaultSettings = {
     theme: "auto",
     enableOTP: false,
     hideBadge: false,
+    nameTemplate: "%host%",
     caps: {
         save: false,
         delete: false,
--- a/src/manifest-firefox.json
+++ b/src/manifest-firefox.json
@@ -29,6 +29,7 @@
         "tabs",
         "clipboardRead",
         "clipboardWrite",
+        "contextualIdentities",
         "nativeMessaging",
         "notifications",
         "scripting",
--- a/src/options/interface.js
+++ b/src/options/interface.js
@@ -59,8 +59,16 @@ function view(ctl, params) {
         createCheckbox.call(this, "enableOTP", "Enable support for OTP tokens (not recommended)")
     );
     nodes.push(createCheckbox.call(this, "hideBadge", "Hide badge counter on the toolbar icon"));
+
     nodes.push(createInput.call(this, "username", "Default username", "john.smith"));
     nodes.push(createInput.call(this, "gpgPath", "Custom gpg binary", "/path/to/gpg"));
+
+    nodes.push(
+        createInput.call(this, "nameTemplate", "Template for new credentials", "%container%/%host%")
+    );
+    nodes.push(
+        m("p", "%host%: tab's hostname; %container%: container name (for Firefox, username otherwise)")
+    );
 
     nodes.push(m("h3", "Theme"));
     nodes.push(
--- a/src/popup/addEditInterface.js
+++ b/src/popup/addEditInterface.js
@@ -284,6 +284,33 @@ function AddEditInterface(settingsModel) {
                 } else {
                     // view instance should be a Login
                     loginObj = new Login(settings);
+                    const template =
+                        helpers.getSetting("nameTemplate", loginObj, settings) || "%host%";
+                    const tab = await chrome.tabs
+                        .query({
+                            lastFocusedWindow: true,
+                            active: true,
+                        })
+                        .then((it) => it[0]);
+
+                    const host = new URL(tab.url).host;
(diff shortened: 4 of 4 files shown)


maximbaz:
Interesting idea! @erayd @patgmiller what are your thoughts on this?

When you say "interface is bad", do you mean the experience of defining such option in the config?

Falling back `%container%` to username feels unnatural to me, maybe we should just leave it empty? :thinking: 

Should we add another placeholder for TLD of the current hostname, i.e. when you are on `account.example.com` to save it as `example.com`?

Not sure that we need `%user%` placeholder altogether - if you defined it in the extension, it's likely because you don't want to define it anywhere in your password store? :thinking:

cab404:
> When you say "interface is bad", do you mean the experience of defining such option in the config?

I mean that I don't like how the explanation in `<p>` turned out looking, but I kind of wanted to push already, since otherwise I might have just continued sitting on it)

> Should we add another placeholder for TLD of the current hostname, i.e. when you are on account.example.com to save it as example.com?

Sounds good

> Falling back `%container%` to username feels unnatural to me, maybe we should just leave it empty? 🤔

Oh, I actually wanted to default to `%host%` — I myself use `%container%/%host%`, but I doubt that containers are that popular

> Not sure that we need %user% placeholder altogether

There is no %user% placeholder — I just use `username` field as a default if %container% is used and missing

E.g I use `identity/site` structure of my password store, and my default identity is `cab/`. But I also have work-related ones.

maximbaz:
> There is no %user% placeholder — I just use username field as a default if %container% is used and missing

Yeah that's what I meant it felt unusual / unexpected for me, why of all the things the fallback is to username, but I see now that you and me just have different things that we express with containers 😅 In any case it's not going to affect you, but I think it's better to not fall back to anything on missing value, let it be missing.

> E.g I use identity/site structure of my password store, and my default identity is cab/. But I also have work-related ones.

By the way you might want to try using multiple password stores for this, you get some colors too and it's kinda nicely follows your structure.

cab404:
> By the way you might want to try using multiple password stores for this, you get some colors too and it's kinda nicely follows your structure.

Multiple password stores as in having multiple separate repositories? That sounds like a lot of hassle when you set up a temporary identity, given you'll have to set it up on all the other devices?

maximbaz:
No you can just map your subfolders in browserpass to be treated as separate password stores, like this:

image

cab404:
> No you can just map your subfolders in browserpass to be treated as separate password stores, like this:
> 
> image
> 
> (could still be a lot of hassle, it's just an idea for you to consider)

that does sound nicer that I thought, I'll try it out!

however I guess that will also require some sort of pairing with containers, so a correct store is chosen when creating a password

maximbaz:
I agree and generally I have to say, it could be cool to support containers in more places - for example have frequency (usage history) respect container.

But let's take it one step at a time, try these custom containers and let me know how it goes and in what direction we should take this PR forward?

cab404:
and immediately I want to add relative paths to stores

patgmiller:
> Interesting idea! @erayd @patgmiller what are your thoughts on this?
> 

I think the idea to allow / provide a personal default path to create secrets is interesting. However I don't think browserpass should set a default value for it. This is one of those things that is subjectively personal for the most part. To me it's like the "default username", 

 Screenshot 2025-05-10 at 10 39 32 AM

however it sounds like maybe with the "container/username/identity" case, the value is configured per store, while the "default username" is global; if that is the case it would require a little more work.

maximbaz:
I agree by the way that there shouldn't be a default value :+1:

patgmiller:
> > By the way you might want to try using multiple password stores for this, you get some colors too and it's kinda nicely follows your structure.
> 
> Multiple password stores as in having multiple separate repositories? That sounds like a lot of hassle when you set up a temporary identity, given you'll have to set it up on all the other devices?

I would say you don't have to setup multiple identities, unless it calls for it. However I understand the multiple repositories, the only hassle for me there is I'd like to have them on my mobile device as well and that's something I've been working towards for a while now. 

Though I do think @maximbaz suggestion to add [multiple paths](https://github.com/browserpass/browserpass-extension/pull/379#issuecomment-2863141679) is a simple solution as well.

cab404:
I guess at this point I only can say that our views on comfort are a bit different, and I will use my fork as daily driver 

However, I can try cutting away at this PR if you want/need any features

max-baz:
Closing for cleanup purposes, we can start fresh PR if we ever want to revisit this - I suppose you are happily using your own fork ever since, and there isn't enough motivation at the moment to bring some of these ideas into this repo.

Issue #377, opened 2025-04-21
Allow to disable browserpass for httpauth


allo-:
I'd like to be able to disable browserpass for HTTP-Auth. Since the last upgrade, which introduced the new dialog for http-auth, it always gets in the way and requires a bit of interaction just to tell browserpass that I do not want it to fill in the password from pass.

I guess browserpass only opens the dialog if there is a password stored for that domain. In my case, it finds a password that has nothing to do with http-auth and probably has no way of knowing that I don't want to use it for http-auth. On the other hand, I still have the http-auth password stored in Firefox and use it from there.

Previously, I had to enter the master password in the first dialog and just hit enter for the second one when I opened the page as one of the first in a new session. Now I get the browser pass window first, which I have to close before I get the other two dialogs.
Also, the browserpass window is floating and not clearly associated with the tab, unlike Firefox's tab-modal login dialog. HTTP-Auth never had the best UX, but the current doorhanger of Firefox itself is quite okay.

Having mixed password stores is not a problem on web pages, as I use the hotkey to insert passwords from browserpass and the context menu of the password field to insert passwords from the Firefox store, so they don't interfere with each other.

I suppose the current extension API doesn't allow to check if Firefox itself has a password stored, and probably doesn't allow to use a similar tab-modal doorhanger dialog, so I think the best way to solve this problem for me would be to have a switch if browserpass should be used for HTTP auth or not.

Related: #374

maximbaz:
Thanks for the additional context! I might be overlooking this, but just to explicitly ask, could you consider an alternative where you'd actually move your credentials into browserpass, and describe how that would look? Besides the initial migration effort, then you'd get just one dialog to choose credentials (the browserpass one), so there won't be _multiple_ dialogs to go through? The fact that browserpass can't bind the floating window to the tab is definitely a point against the built-in FF password manager, but besides that, is there any downside to migrating your passwords to browserpass, since you are using this extension anyway?

allo-:
I'm not quite sure. I have currently a wild mix, of which some passwords are migrated one by one to pass, but in the end there are two use-cases which work quite well by having the default unlocked Firefox password safe and the more secure pass store.

[...]

PR #381, opened 2025-05-17
handle url filter correctly when host is not a valid domain

patgmiller:
handle url filter correctly when host is not a valid domain, fall back to host or empty string

fixes bug identified in browserpass/browserpass-extension#377

### DIFF
--- a/src/popup/interface.js
+++ b/src/popup/interface.js
@@ -234,7 +234,7 @@ function search(searchQuery) {
 
     if (overrideDefaultSearchOnce && this.settings.authRequested && authUrl) {
         const authUrlInfo = new BrowserpassURL(authUrl);
-        searchQuery = authUrlInfo?.validDomain ? authUrlInfo.domain : authUrlInfo?.host || "";
+        searchQuery = authUrlInfo.validDomain ? authUrlInfo.domain : authUrlInfo.hostname ?? "";
         this.results = helpers.filterSortLogins(this.logins, searchQuery, true);
     } else {
         this.results = helpers.filterSortLogins(this.logins, searchQuery, this.currentDomainOnly);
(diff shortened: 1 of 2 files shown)


patgmiller:
@maximbaz @erayd I confirmed this fix also works for Firefox, in addition to Chrome, since FF is where it was originally identified in #377 

[loading-avail-logins-fixed-ff-chrome.webm](https://github.com/user-attachments/assets/07f9740e-b30b-4455-b3dd-c09050a1e386)

src/popup/interface.js
```diff
@@ -234,7 +234,8 @@ function search(searchQuery) {
 
     if (overrideDefaultSearchOnce && this.settings.authRequested && authUrl) {
         const authUrlInfo = new BrowserpassURL(authUrl);
-        this.results = helpers.filterSortLogins(this.logins, authUrlInfo.domain, true);
+        searchQuery = authUrlInfo?.validDomain ? authUrlInfo.domain : authUrlInfo?.host || "";
```
maximbaz, line 237:
```suggestion
        searchQuery = authUrlInfo.validDomain ? authUrlInfo.domain : authUrlInfo.hostname ?? "";
```

Please correct me if I'm wrong, but since we pass `domain` here, domain doesn't include the port number, so for consistency sake we should fallback to `hostname` and not `host` to also not include the port number.

We do have this in the README, I hope it still holds true :sweat_smile: 

> Browserpass can also distinguish credentials meant for different ports, so for example an entry example.com.gpg will show up in Browserpass popup when you browse example.com on any port, however an entry example.com:8080.gpg will only show up on 8080 port.

Other than that, I suppose authUrlInfo can never be null/undefined, so we can drop the `?` right? (and use `??` for minor preference)

patgmiller:
@maximbaz nice catch, that's one I was not aware of; i had to make a second correction b/c I used the host for the search label as well.

maximbaz (approved):

PR #386, opened 2025-05-29
folder completion: preserve input rhs of cursor

maximbaz:
Suppose your password store has folders "personal" and "work".
Press "Add credentials" in the popup, in the file name enter "github.com" (will be automatically done by #383), then press Home and try to autocomplete "personal" folder.

Before: autocomplete wouldn't show up
After: autocomplete shows up based on what there is on the left side of the cursor. So you can type "p" and it will exclude "work" and keep showing "personal". Pressing Enter prepends folder name, without overwriting "github.com".

### DIFF
--- a/src/popup/addEditInterface.js
+++ b/src/popup/addEditInterface.js
@@ -91,13 +91,7 @@ function AddEditInterface(settingsModel) {
                 case "Enter":
                     e.preventDefault();
                     if (e.target.classList.contains("directory")) {
-                        // replace search term with selected directory
-                        inputEl.value = `${addDirToLoginPath(
-                            loginObj.login,
-                            e.target.getAttribute("value")
-                        )}/`;
-                        this.state.setLogin(inputEl.value);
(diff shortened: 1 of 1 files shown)


src/popup/addEditInterface.js
```diff
@@ -143,9 +137,12 @@ function AddEditInterface(settingsModel) {
         function clickDirectoryHandler(e) {
             e.preventDefault();
             var inputEl = document.querySelector("input.filePath");
-
-            // replace search term with selected directory
-            inputEl.value = `${addDirToLoginPath(loginObj.login, e.target.getAttribute("value"))}/`;
+            var dir = e.target.getAttribute("value");
```
patgmiller, line 140: This is purely a nitpick, so feel free to disregard if you wish. But since these variables are never updated I would suggest using `const`.

patgmiller (approved):
@maximbaz LGTM! I let one extremely minor suggestion which is optional.

Issue #404, opened 2026-05-15
TypeError: Cannot read properties of null (reading 'status')


bersace:
Hi,

I hit errors with:

- Brave 1.90.122
- extension: 3.11 (from store)
- native: 3.1.2 (manual installation)

Password count is ok.

When opening popup, I hit #394 

When opening extensions options: 

<img width="685" height="286" alt="Image" src="https://github.com/user-attachments/assets/922e240c-55e2-4f0a-9949-53136751370f" />

I don't know what happen.

bersace:
Reinstalling *web* extension fixed it.

bersace:
Reopening since it's happening again. Is it an issue with latest Brave ?

bersace:
Disabling-reenabling extension, workaround for a moment.

max-baz:
With https://github.com/browserpass/browserpass-extension/commit/dbf30abcc138a4b6fec2f849b32b1ec6c8bd4a9b in `master`, the extension shouldn't at least crash anymore - give it a try. It isn't going to fix the root cause, as we don't really know what's happening. If you still experience errors, it would be helpful if you could try to collect more information, to make it actionable for us. Thanks!

bersace:
@max-baz thanks ! Do you know where to catch extension errors ?

max-baz:
chrome://extensions/ > Browserpass

On a crash there typically will be a button to see error details and stack trace, and also there's "Inspect views service worker" link that can open devtools, where you can maybe get some more hints. Good luck!

fblaese:
I am running into the same issue. Sadly, dbf30abcc138a4b6fec2f849b32b1ec6c8bd4a9b has not fixed it. I don't see any other errors in the chromium extension page or in the inspector console for both the service worker and popup, apart from `Cannot read properties of null (reading 'status')`. The UI is not able to read the settings (`getFullSettings()`) anymore after some time (both in the popup and the extension settings), until I disable and re-enable the extension.

According to chrome://serviceworker-internals/, the service worker has `Running Status: RUNNING`.

I had the same issue a while back (~1/2 year ago). Back then, I was able to resolve it by completely uninstalling and reinstalling the chromium extension.

Do you have any suggestions on how to debug this further? Is it possible that this is a chromium bug afterall? Because I am able to set break points in `background.dist.js` at `handleMessage()`, which are triggered and the function seems to behave correctly. But the response never reaches the popup.

max-baz:
[...]

PR #406, opened 2026-05-22
ensure addListener takes a synchronous callback

max-baz:
Another potential improvement to help with #404 

As per https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#sendresponse:

image

I'm reusing the same pattern used in:

https://github.com/browserpass/browserpass-extension/blob/0138231cd1828726566eb3c9132aa179c72c4793/src/background.js#L61-L67

### DIFF
--- a/src/offscreen/offscreen.js
+++ b/src/offscreen/offscreen.js
@@ -3,7 +3,12 @@
 const clipboard = require("../helpers/clipboard");
 
 //----------------------------------- Function definitions ----------------------------------//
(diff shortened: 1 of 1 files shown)


Nik-mmzd:
Hello! Can you compile and publish the extension as pre-release for those who are not able to compile extension themselves or want signed crx?

max-baz:
Done :heavy_check_mark:

